PR #9, opened 2017-05-11
Dependencies fix

### LINKED ISSUE #8: npm dependencies
Hi, I've noticed that dependencies in `package.json` are a bit mixed up: `@types` are in `dependencies` and regular libraries that are used are in `dev-dependencies`.

The thing is that npm won't install `dev-dependencies` when you install the package, but only `dependencies`. So if you say that someone needs to install `reflect-metadata` when they install this package, that's an obvious "smell" you need it in your `dependencies`.

I'll make a PR soon with a fix 🙂

Pritilender:
PR to fix #8 
I'm not sure how to test this.

Also, I've noticed that there is also a dependency for sequelize package note in readme. Wouldn't it be better that this package be a standalone sequelize TS wrapper, and that it should include sequelize by default? Just a thought.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,13 +30,6 @@ Decorators and some other extras for sequelize (v3 + v4).
 npm install sequelize --save // v3
 npm install sequelize@4.0.0-1 --save // or v4
 ```
-and [reflect-metadata](https://www.npmjs.com/package/reflect-metadata)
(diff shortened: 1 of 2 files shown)


RobinBuschmann:
see #8

PR #19, opened 2017-06-06
Bumped @types/sequelize

snewell92:
`@types/sequelize` got an update, and I let npm [generate the package-lock.json](https://docs.npmjs.com/files/package-lock.json) file that npm v5 throws in the root of the project now.

I tried running the tests, but I got a weird error.
```
> SEQ_SILENT=true mocha 'test/specs/**/*.js'                      
                                                                  
'SEQ_SILENT' is not recognized as an internal or external command,
operable program or batch file.                                   
```

Windows 10 64bit
Node.js: **8.0.0**
NPM: **v5.0.2**
Typescript: **v2.3.2**

I want to see if the test passes on travis, then it's probably a windows/environment thing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,11 @@
       "integrity": "sha1-EhNEEGcsFkO9qlrKIzmw43F68Yg=",
       "dev": true
     },
+    "@types/geojson": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@types/geojson/-/geojson-1.0.2.tgz",
+      "integrity": "sha1-sC0QqwKOKSisWSoFGqpJgaGUHQM="
+    },
     "@types/lodash": {
       "version": "4.14.54",
       "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.54.tgz",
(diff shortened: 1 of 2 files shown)


snewell92:
Tests passed on travis, but fail on my local machine. 😢 Still getting same error. Differences in OS and node/npm/typescript are likely the cause...? Maybe on windows [we need something like this](https://stackoverflow.com/a/9250168/2378943)...?

Adding a pretest and precover script and removing the environment variable declarations in `cover-v*` and `exec-tests` could work.

**package.json** -> scripts:
```
... other scripts...
"pretest": "set SEQ_SILENT=true",
"precover": "set SEQ_SILENT=true",
... other scripts...
```

Or we could add another piece to the prepare-test-v* scripts.

Or we could remove the SEQ_SILENT altogether and just have `.travis.yml` set that environment variable before running tests. *I like this the best*.

@RobinBuschmann what do you think? I have access to both a windows and ubuntu testing environment and could test changes to the scripts to make sure the npm scripts all run successfully both on windows and ubuntu, but as it is the tests pass on Travis, and that could be good enough for now I suppose.

snewell92:
Oh the changes I actually want are in `v4.0.59` on the sequelize types.

RobinBuschmann:
@snewell92 putting the SEQ_SILENT flag to travis.yaml sounds like a good idea. :)

Regarding your pull request: You've accidently added sequelize 4.0.0 to the final dependencies.

snewell92:
Whoops! Proof I ran the tests I guess xD fix inc...

snewell92:
I pulled the SEQ_SILENT stuff into its own issue and branch that I started work on. Ran into a problem with `nyc` in the npm scripts, detailed over on the referenced issue.

PR #25, opened 2017-06-13
Filter out ts definitions

### LINKED ISSUE #23: Model loading, ignore declaration files
When passing a directory to the `modelPaths` option in the constructor, it would be nice if sequelize-typescript only looked for .js files.

I see two instances where people could run into this:

1. `mocha-typescript` requires my main application to be compiled with the `"declarations": true` option so that it can consume the application's types to (if the test suite and app are two different typescript apps). Unfortunately, this means in my output directory not only will typescript emit `js` files but also `.d.ts`. Anyone else using **declarations** will run into this.

2. Any source maps **map** file alongside the models will also junk up the output directory.

This can be easily worked around with a little but of code with `fs` and some filtering, but maybe this extra nicety could be baked into sequelize-typescript? I'm going to be writing up a PR, since I'll be using the same logic in my own code if this doesn't make it into the repo. Feel free to either close this issue if you don't want it baked in, or assign it to me if you do.

Thanks for reading! <3<3

snewell92:
Only adds filter predicate to test against the `.d.ts` extension. It is also worth mentioning, although not applicable currently, that the `extension` local variable in `models.ts` isn't necessarily the extension if the directory has files with longer extensions (for example, the .d.ts extension needs that splice(-5)). Just a minor nit picky thing, can fix later (we're not expecting any other file types other than .js/.ts/.d.ts).

Added a comment/TODO for it for future reference.

Closes #23

### DIFF
--- a/lib/services/models.ts
+++ b/lib/services/models.ts
@@ -170,8 +170,9 @@ export function getModels(arg: Array<typeof Model|string>): Array<typeof Model> 
       const _models = fs
         .readdirSync(dir as string)
         .filter(file => {
+          // TODO extension is not necessarily only the lasst three characters
           const extension = file.slice(-3);
-          return extension === '.js' || (extension === '.ts' && file.slice(-4) !== '.d.ts');
+          return extension === '.js' || (extension === '.ts' && file.slice(-5) !== '.d.ts');
         })
         .map(file => path.parse(file).name)
         .filter(uniqueFilter)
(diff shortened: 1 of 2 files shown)


RobinBuschmann:
Thanks for fixing. But codecov is not happy, although I don't really understand why :/ What does "50% of diff hit (target 94.5%)" mean?

RobinBuschmann:
Ah ok, 50% of your changes are not covered. Since you accidently removed a whitespace from a part of the code, that was not even covered in the past, codecov blames you for that :P But I think this should be ok. I'll merge it back into master :)

snewell92:
I think this might be because of an editor config or something... Weird. I'll be more diligent in examining my diffs before committing. No big deal this time. :D

Issue #37, opened 2017-06-22
Model.scope().findAll({...}) cannot infer `as` in `include` option


RobinBuschmann:
Assuming the following defined models:
```typescript
class User extends Model<User> {
  companies: Company[];
}

class CompanyUser extends Model<CompanyUser> {
 @ForeignKey(() => Company) 
 @Column
 companyId: number;

 @ForeignKey(() => User) 
 @Column
 userId: number;
}

@Scopes({bla: {...}})
class Company extends Model<Company> {
  users: User[];
}
```

When selecting entities of a scoped model, the `as` value of the include options, that are NOT defined within the scope, is not inferred by sequelize-typescript:

```typescript
Company
   .scope('bla')
   .findAll({include: [User]})
  .catch(...) // throws "User is not associated to Company!"
;
```

Current workaround:
```typescript
Company
   .scope('bla')
   .findAll({include: [{
     model: User,
     as: 'users' // <-- this need to be set explicitly
   }]})
  .catch(...)
;
```

RobinBuschmann:
Same issue for `findAndCountAll()`

PR #42, opened 2017-07-06
allow additional options to be specified for associations

### LINKED ISSUE #34: Provide association options
I'm trying to set some association options. For example, I have a `HasMany` where I want to disallow null in the FK and cascade on delete.

In Sequelize, I'd do something like:

```
childModel.belongsTo(parentModel, {
   foreignKey: {
      allowNull: false
   },
  onDelete: 'CASCADE'
});
```

I don't see any place in the association attributes or `ForeignKey` to pass in those options. See [foreign keys](http://docs.sequelizejs.com/manual/tutorial/associations.html#foreign-keys) (Their documentation on this subject is a bit messy).

I'd like to be able to do the following, using BelongsTo as an example:
```
export class Child extends Model<Child> {
   @BelongsTo(() => Parent, { foreignKey: { allowNull: false }, onDelete: 'CASCADE'})
    public parent: Parent;
}
```
I'm using Sequelize 4, but I think the issue is the same for 3.

HackedByChinese:
Affects `BelongsTo`,  `BelongsToMany`, `HasMany`, `HasOne`, and `ForeignKey` attributes. Compatibility with existing parameter that takes string name of foreign keys has been provided by letting the foreign key be a string (existing functionality) or the appropriate association options (new functionality). Additionally, `ForeignKey` can now receive options. By default it simply sets the foreign key name to the property name, which should be backwards compatible.

I did not add tests for the Many-to-Many cases, because additional options outside setting the FK column names and table name are mostly superfluous.

Fixes #34

### DIFF
--- a/lib/annotations/ForeignKey.ts
+++ b/lib/annotations/ForeignKey.ts
@@ -1,10 +1,17 @@
+import {AssociationForeignKeyOptions} from 'sequelize';
 import {Model} from "../models/Model";
 import {addForeignKey} from "../services/association";
 
-export function ForeignKey(relatedClassGetter: () => typeof Model): Function {
+export function ForeignKey(relatedClassGetter: () => typeof Model, options?: AssociationForeignKeyOptions): Function {
 
   return (target: any, propertyName: string) => {
 
-    addForeignKey(target, relatedClassGetter, propertyName);
+    if (!options) {
+      options = {name: propertyName};
+    } else if (!options.name) {
+      options.name = propertyName;
+    }
+
+    addForeignKey(target, relatedClassGetter, options);
   };
 }
--- a/lib/annotations/association/BelongsTo.ts
+++ b/lib/annotations/association/BelongsTo.ts
@@ -1,17 +1,23 @@
+import {AssociationOptionsBelongsTo} from 'sequelize';
+
 import {Model} from "../../models/Model";
 import {BELONGS_TO, addAssociation} from "../../services/association";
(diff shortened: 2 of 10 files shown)


RobinBuschmann:
@HackedByChinese thanks for implementing this. Looks good.

> codecov/patch — 92.85% of diff hit (target 94.5%)

is not an issue. I will merge it anyway

HackedByChinese:
Hello! Any comment on when we might see this in a release? Thanks!

RobinBuschmann:
Hey, I'm planning to release a new version at the end of this weekend or not later than next week. To make this happen I need to fix #43 and #37 .

RobinBuschmann:
@HackedByChinese Hey, I removed the AssociationForeignKeyOptions  from the `ForeignKey` annotation, because these options only consists of a name and `ColumnOptions`. These options should rather be set via `Column` annotation, which is already part of sequelize-typescript.

So that the following example, which shows your implementation..
```typescript
@ForeignKey(() => Model, {name: 'fooId', field: 'foo_id'})
fooId: number;
```
..should be achieved with
```typescript
@ForeignKey(() => Model)
@Column({field: 'foo_id'})
fooId: number;
```
I think that this approach makes the api much cleaner. What do you think?

Additionally I added support for `otherKey` in the AssociationOptions of `BelongsToMany`.

HackedByChinese:
Sure, that makes sense. Thanks!

PR #47, opened 2017-07-10
transaction is allowed in association options

imposibrus:


RobinBuschmann:
Thanks for contributing

PR #63, opened 2017-07-22
Fixing findCreateFind to return [Model, created]

kingjerod:
This is what Sequelize returns.


kingjerod:
Opened a new PR that includes this one and a few other fixes.

PR #64, opened 2017-07-22
Fixing findOrCreate and findCreateFind

kingjerod:
1) Fixing findOrCreate to allow any values or a specific interface A. This matches the findOrBuild definition.

2) Fixing findCreateFind to return the correct value [Model, boolean], and also apply the same fix as #1.


RobinBuschmann:
@kingjerod Looks good, thanks for fixing!

PR #71, opened 2017-07-28
Don’t force the table name

natesilva:
If you don’t specify a table name, `sequelize-typescript` uses the class name. This is different from what Sequelize does. One example: for a model `Person`, Sequelize creates the table `People`. But `sequelize-typescript` creates `Person` instead.

By default Sequelize pluralizes table names. There are other options that affect the table name including `underscoredAll` and `freezeTableName`.

By simply not trying to set the table name, you would get the default behavior for free. As far as I can tell, `sequelize-typescript` has no dependency on the specific table name.

This would be a major breaking change, of course, because every model’s table name would change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,8 @@ class Person extends Model<Person> {}
 
 Decorator                             | Description
 --------------------------------------|---------------------
- `@Table`                             | sets `options.tableName=<CLASS_NAME>` and  `options.modelName=<CLASS_NAME>` automatically
- `@Table(options: DefineOptions)`     | sets [define options](http://docs.sequelizejs.com/manual/tutorial/models-definition.html#configuration) (also sets `options.tableName=<CLASS_NAME>` and  `options.modelName=<CLASS_NAME>` if not already defined by define options) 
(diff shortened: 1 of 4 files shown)


RobinBuschmann:
Hey @natesilva thanks for your PR. It was a conscious decision to use the class name as tablename. Probably not a good one. As you noticed it differs from the behaviour of sequelize. A better approach would had been, if I achieved this behaviour via a global option passed to an instance of `Sequelize`. For example `useClassNameAsTableName: true`.
Your PR would mean a major change now and would break projects, that are using sequelize-typescript. I would suggest to pick up the idea of a global option to achieve the old behaviour of sequelize and preventing sequelize-typescript from using the class name as table name automatically (`preventInferTableName: true` or the like). What do you think?

natesilva:
Thanks for taking a look; I figured it was too disruptive to merge, but appreciate the explanation. Coming from plain Sequelize, I spent a few minutes trying to figure out why my table names were wrong. It would be helpful to have an option like `preventInferTableName` for those of us migrating. For now we’ll just add `tableName` to each of our `@Table` declarations.

Issue #73, opened 2017-07-29
Add connection string initialization


aventurella:
Per the docs:
http://docs.sequelizejs.com/manual/installation/getting-started.html#setting-up-a-connection

This should be allowed to be passed into the `Sequelize` constructor:

```js
// Or you can simply use a connection uri
const sequelize = new Sequelize('postgres://user:pass@example.com:5432/dbname');
```

https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L103-L134

This appears to work (I ported from their code):

```js
function parseConnectionString(connectionString: String): any{
    let config: any = {};
    let options: any = {}

    const urlParts = url.parse(connectionString);

    options.dialect = urlParts.protocol.replace(/:$/, '');
    options.host = urlParts.hostname;

    if (options.dialect === 'sqlite' && urlParts.pathname && urlParts.pathname.indexOf('/:memory') !== 0) {
        const path = Path.join(options.host, urlParts.pathname);
        options.storage = options.storage || path;
    }

    if (urlParts.pathname) {
        config.database = urlParts.pathname.replace(/^\//, '');
    }

    if (urlParts.port) {
        options.port = urlParts.port;
    }

    if (urlParts.auth) {
        const authParts = urlParts.auth.split(':');

        config.username = authParts[0];

        if (authParts.length > 1)
            config.password = authParts.slice(1).join(':');
    }

    let result = Object.assign({}, config, options);

    return result;
}
```

hiradimir:
+1

But, now I am using "parse-database-url".


```ts
import * as parseDbUrl from "parse-database-url";
---
      let dbConfig = parseDbUrl(process.env["DATABASE_URL"]);

      const sequelize = new Sequelize({
        name: dbConfig.database,
        username: dbConfig.user,
        password: dbConfig.password,
        dialect: dbConfig.driver,
        host: dbConfig.host,
        port: dbConfig.port,
        modelPaths: [__dirname + '/../main/models']
      });

```

aventurella:
Winning! I'll take that all day. Personal preference for simple stuff like that which is only a few LOC i'm dependency averse. At the same time, whatever works ;)

RobinBuschmann:
@aventurella thanks for raising this. I will consider this for the next releases. But if you're like to create a PR for that, feel free to do so. Any help is appreciated :)

kukoo1:
[...]

PR #79, opened 2017-07-31
connect using URI connection string support

kukoo1:
As mentioned in #73.
Typescript doesn't allow spreading arguments into function [https://github.com/Microsoft/TypeScript/issues/4130](https://github.com/Microsoft/TypeScript/issues/4130) and doesn't allow conditional super() call [https://github.com/Microsoft/TypeScript/issues/945](https://github.com/Microsoft/TypeScript/issues/945).

One way to achieve this is to instantiate Sequelize in a static factory function.
Old constructor must be removed to allow arguments to be exactly relayed to OriginSequelize constructor. Thus, this change will break backward compatability.

Developers currently using "new Sequelize(config)" have to change the code to "Sequelize.init(config)" or "Sequelize.init(URI, config)".

Should this be merged into next major version ?

### DIFF
--- a/lib/annotations/association/BelongsToMany.ts
+++ b/lib/annotations/association/BelongsToMany.ts
@@ -20,7 +20,7 @@ export function BelongsToMany(relatedClassGetter: ModelClassGetter,
     through = throughOrOptions;
   } else {
     through = (throughOrOptions as IAssociationOptionsBelongsToMany).through;
-    options = throughOrOptions;
+    options = throughOrOptions as IAssociationOptionsBelongsToMany;
   }
   return (target: any, propertyName: string) => {
     addAssociation(
--- a/lib/annotations/validation/IsUUID.ts
+++ b/lib/annotations/validation/IsUUID.ts
@@ -1,10 +1,12 @@
 import 'reflect-metadata';
 import {addAttributeOptions} from "../../services/models";
 
-/**
- * Only allow uuids
+/*
+ * Only allow uuids.
+ * Version's regular expressions:
+ * https://github.com/chriso/validator.js/blob/b59133b1727b6af355b403a9a97a19226cceb34b/lib/isUUID.js#L14-L19.
  */
-export function IsUUID(version: number): Function {
+export function IsUUID(version: 3|4|5|"3"|"4"|"5"|"all"): Function {
 
   return (target: any, propertyName: string) =>
     addAttributeOptions(target, propertyName, {
--- a/lib/interfaces/IBaseIncludeOptions.ts
+++ b/lib/interfaces/IBaseIncludeOptions.ts
@@ -15,7 +15,7 @@ export interface IBaseIncludeOptions {
    * Where clauses to apply to the child models. Note that this converts the eager load to an inner join,
    * unless you explicitly set `required: false`
    */
-  where?: WhereOptions;
+  where?: WhereOptions<any>;
 
   /**
    * A list of attributes to select from the child model
--- a/lib/interfaces/IDefineOptions.ts
+++ b/lib/interfaces/IDefineOptions.ts
@@ -4,7 +4,9 @@ export interface IDefineOptions extends DefineOptions<any> {
   modelName?: string;
 
   /**
-   * To enable optimistic locking.
+   * Enable optimistic locking.  When enabled, sequelize will add a version count attribute
+   * to the model and throw an OptimisticLockingError error when stale instances are saved.
+   * Set to true or a string with the attribute name you want to use to enable.
    */
-  version?: boolean;
+  version?: boolean | string;
 }
--- a/lib/interfaces/IFindOptions.ts
+++ b/lib/interfaces/IFindOptions.ts
@@ -11,7 +11,7 @@ export interface IFindOptions extends LoggingOptions, SearchPathOptions {
   /**
    * A hash of attributes to describe your search. See above for examples.
    */
-  where?: WhereOptions | Array<col | and | or | string> | col | and | or | string;
+  where?: WhereOptions<any> | Array<col | and | or | string> | col | and | or | string;
 
   /**
    * A list of the attributes that you want to select. To rename an attribute, you can pass an array, with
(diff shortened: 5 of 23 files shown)


RobinBuschmann:
Hey @kukoo1 thanks for contributing :)

Why is spreading necessary here, from your point of view?

Did you try to resolve this with overloading the constructor like:
```typescript
export class Example {

  constructor(connection: string);
  constructor(options: Options);
  constructor(optionsOrConnection: Options|string) {
    /*...*/
  }
}
```
?

kukoo1:
I've tried this.
```javascript
constructor(config: ISequelizeConfig, uri?: string) {
    // Sequelize constructor signature:
    //// new (database: string, username: string, password: string, options?: Options): Sequelize;
    //// new (database: string, username: string, options?: Options): Sequelize;
    //// new (uri: string, options?: Options): Sequelize;
   if (uri) {
      super(uri, BaseSequelize.prepareConfig(config));
   } else {
      super(
         (preparedConfig = BaseSequelize.prepareConfig(config), preparedConfig.name),
         preparedConfig.username,
         preparedConfig.password,
         preparedConfig
      );
   }
   this.init(config);
}
```
The compiler complains
`A 'super' call must be the first statement in the constructor when a class contains initialized properties`

So I tried solving with this.
```javascript
constructor(config: ISequelizeConfig, uri?: string) {
   super(
      uri || (preparedConfig = BaseSequelize.prepareConfig(config), preparedConfig.name),
      uri ? BaseSequelize.prepareConfig(config) : preparedConfig.username,
      uri ? undefined : preparedConfig.password,
      uri ? undefined : preparedConfig
    );
    this.init(config);
}
```
Now it works with traditional connection options as an object.
But by using connection string, it won't work.
Because originalSequelize will recognize that super(uri, config, undefined, undefined) contains 4 arguments, so "uri" would be db name, "config" would be username, "undefined" would be password and so on.

Any idea to this? :)

RobinBuschmann:
@kukoo1 Thank you for your input :)

Or much simpler:
```typescript
new Sequelize({uri: '...'})
```
```typescript
constructor(config: ISequelizeConfig | ISequelizeUriConfig) {}
```
Since sequelize-typescript already uses one object literal for options instead of multiple parameters, I think this will be the most consistent approach. What do you think?

BruceHem:
hmm it's just a string, do we really need to make it object literal?

RobinBuschmann:
@BruceHem I think it is most likely, that the configuration will look like this

```typescript
new Sequelize({
  uri: '...',
  modelPaths: ['...']
})
``` 

But in addition, we could overload the constructor like:

```typescript
  constructor(connection: string);
  constructor(options: ISequelizeConfig | ISequelizeUriConfig);
  constructor(optionsOrConnection: ISequelizeConfig | ISequelizeUriConfig | string) {
    /*...*/
  }
```

kukoo1:
I found the root of this mess.
I've just looked into Sequelize code.
It accepts
```javascript
new Sequelize({ ... options })
new Sequelize(URI, { ... options })
new Sequelize(database, username, password, { ... options })
```
So we could call super(uriOrOption) properly.

At first, when I coded this PR, I only looked at Sequelize.d.ts definition (@types/sequelize).
The definition accepts only
```javascript
new (uri: string, options?: Options)
new (database: string, username: string, options?: Options)
new (database: string, username: string, password: string, options?: Options)
```

So we've to open PR in DefinitelyTyped to add this beforehand
```javascript
new (options: Options): Sequelize;
```

kukoo1:
Now I get into another trouble. I can't pass both URI and options at the same time.
current constructor looks like this
```javascript
constructor(configOrUri: ISequelizeConfig | ISequelizeUriConfig | string) {
   super(
      (typeof configOrUri === "string") ?
        configOrUri : // URI connection string
        BaseSequelize.prepareConfig(configOrUri) // config object
   );
}
```
Sequelize checks arguments using this code
```javascript
if (arguments.length === 1 && typeof database === 'object') {
   // new Sequelize({ ... options }) <-- in order THIS to WORKS, we must ONLY pass 1 argument !!
   // ...
} else if (arguments.length === 1 && typeof database === 'string' || arguments.length === 2 && typeof username === 'object') {
   // new Sequelize(URI, { ... options })
   // ...
}  else {  // <-- with this, URI string will NOT WORKING !!
   // new Sequelize(database, username, password, { ... options })
   // ...
}
```
So we couldn't do this.
```javascript
super(
   (typeof configOrUri === "string") ?
      configOrUri : // URI connection string
      BaseSequelize.prepareConfig(configOrUri), // config object
   (configOrUri.uri) ?
      configOrUri : // ISequelizeUriConfig
      undefined // <-- THIS will BREAK arguments.length === 1
);
```
There're 2 options.
1) ignore config when URI is passed (ISequelizeUriConfig will be a junk and we'll lose flexibility when use URI connection string)
2) open PR in Sequelize to add `const argLength = arguments.filter(val => val !== undefined).length` and use `if (argLength === 1)` instead.

RobinBuschmann:
@kukoo1 Ahh ok, I see. I didn't get it in the first place.

From my understanding, typescript makes it impossible to use overloads in constructors in combination with inheritance in general. Cause it is very common to check arguments length. Or have we overseen something?

kukoo1:
waiting for review ...
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18896

kukoo1:
@RobinBuschmann  I committed the new one.
```javascript
constructor(config: ISequelizeConfig | ISequelizeUriConfig | ISequelizeDbNameConfig);
constructor(uri: string);
```

I suggest to deprecate "name" property in options, and use "database" instead.
Because, in Sequelize documentation, "database" is used.
In order to use the same config when switching from ordinary sequelize, we should use "database".
What's your opinion ?

RobinBuschmann:
Looks good to me. Deprecating "name"  is reasonable as well. But instead of adding "@Todo", you could write "@deprecated" - this will be noticed by many IDEs and therefore be notice the user that "name" is deprecated.

RobinBuschmann:
Fixes #73

PR #86, opened 2017-08-06
OptimisticLocking property: number --> number|string

IzikLisbon:
http://docs.sequelizejs.com/manual/tutorial/models-definition.html

```
// Enable optimistic locking.  When enabled, sequelize will add a version count attribute
// to the model and throw an OptimisticLockingError error when stale instances are saved.
// Set to true or a string with the attribute name you want to use to enable.
```

IDefineOptions: as commented above -`Set to true or a string`.
Model.d.ts: if a user sets the version in IDefineOptions to a string (or not set it at all) it is probably to free the version field for something else and in that case it should not be forced to a number (This is the motivation for this pull request - I had a version field of type string which called a build break).

### DIFF
--- a/lib/interfaces/IDefineOptions.ts
+++ b/lib/interfaces/IDefineOptions.ts
@@ -4,7 +4,9 @@ export interface IDefineOptions extends DefineOptions<any> {
   modelName?: string;
 
   /**
-   * To enable optimistic locking.
+   * Enable optimistic locking.  When enabled, sequelize will add a version count attribute
+   * to the model and throw an OptimisticLockingError error when stale instances are saved.
+   * Set to true or a string with the attribute name you want to use to enable.
    */
-  version?: boolean;
+  version?: boolean | string;
 }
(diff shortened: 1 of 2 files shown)


RobinBuschmann:
@IzikLisbon Hey, thanks for contributing. What does "something else" mean for Mode.d typings? What else could it be than a `boolean`, `string` or a `number`? Thank you

IzikLisbon:
@RobinBuschmann  It can be whatever the user decides it to be. Once the user is setting `version` in `IDefineOption` to a string, the `version` in `Mode.d.ts` can be of any type.  Most likely that it will be a `boolean`, a `string` or a `number` but there is no reason to force it.

RobinBuschmann:
Fair enough. Thanks again.

Jendorski:
I cannot find `IDefineOptions` in sequelize-typescript ver ^2..1.3

`import { IDefineOptions } from 'sequelize-typescript/lib/interfaces/IDefineOptions';`

PR #92, opened 2017-08-10
Add support for all UUID versions 

IzikLisbon:
Sequelize is using Validator to validates UUID strings.
Validator supports 4 versions of UUID - 3,4,5 and all:
https://github.com/chriso/validator.js/blob/b59133b1727b6af355b403a9a97a19226cceb34b/lib/isUUID.js#L14-L19

This change is updating IsUUID to support all possible options - 3,4,5 and all.

see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18777#issuecomment-321458784

**Other stuff**
Two additional changes:
1. `WhereOptions` --> `WhereOptions<any>` 
2. added `as IAssociationOptionsBelongsToMany`

I am not sure about any of those additional changes. I know it passed compilation and testing, but I basically choose the option that compiles. 

Also running `npm run build` before making any change failed.  I have upgraded to node 8 and then it passed. Is there any dependency on the node version or it was just something locally on my machine?

### DIFF
--- a/lib/annotations/association/BelongsToMany.ts
+++ b/lib/annotations/association/BelongsToMany.ts
@@ -20,7 +20,7 @@ export function BelongsToMany(relatedClassGetter: ModelClassGetter,
     through = throughOrOptions;
   } else {
     through = (throughOrOptions as IAssociationOptionsBelongsToMany).through;
-    options = throughOrOptions;
+    options = throughOrOptions as IAssociationOptionsBelongsToMany;
   }
   return (target: any, propertyName: string) => {
     addAssociation(
--- a/lib/annotations/validation/IsUUID.ts
+++ b/lib/annotations/validation/IsUUID.ts
@@ -1,10 +1,12 @@
 import 'reflect-metadata';
 import {addAttributeOptions} from "../../services/models";
 
-/**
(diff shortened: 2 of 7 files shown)


RobinBuschmann:
Looks good, thank you for implementing that.

One question: You've updated `@types/sequelize` to  `4.0.68`. Was this a required step to achieve this feature?

IzikLisbon:
@RobinBuschmann yes. It needed this change: 
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18777

IzikLisbon:
@RobinBuschmann build and tests have failed on my machine. I tried the following
1) clone 
2) `npm install`
3) `npm run build`

Got a failure (Don't remember what it was). 
Upgraded to node8, ran again. This time it was successful. Is there any dependency on the node version?

Then I have tried to run:
`npm run cover`

got an error, asking to install sqlite3 manually. So I ran:
1) `npm install sqlite3`
2) `npm run cover`

This time it worked.

I can open an issue for this, but wanna know if you have any ideas.

IzikLisbon:
@RobinBuschmann do you think this can be pushed?

PR #97, opened 2017-08-14
feature(hooks): add decorators for declaring hooks

### LINKED ISSUE #88: A `@Hook` decorator would be great
There are several ways to add hooks, but they need your `Model` to be added to a Sequelize instance first. It isn’t very declarative.

It would be great if there was a `@Hook` decorator. Something like:

```typescript
@Table
class Book extends Model<Book> {
  // (bunch of @Columns etc.)

  @Hook('beforeCreate')
  static reticulateSplines(instance: Book) {
    // do beforeCreate hook stuff here
  }

  @Hook('afterDestroy')
  static herdLlamas(instance: Book) {
    // do afterDestroy hook stuff here
  }
}
```

Is this in the works? I could try to write something. The decorator would store information using the `reflect-metadata` system and then something like `initializeHooks` would be called from the `addModels` method.

natesilva:
Decorators for declaring hook methods in Model classes.

Includes README update, tests (100% coverage of new code).

Fixes #88

### DIFF
--- a/lib/annotations/hooks/AfterBulkCreate.ts
+++ b/lib/annotations/hooks/AfterBulkCreate.ts
@@ -1,21 +1,8 @@
 import {IHookOptions} from "../../interfaces/IHookOptions";
-import {addHook} from "../../services/hooks";
+import {implementHookDecorator} from "../../services/hooks";
 
 export function AfterBulkCreate(target: any, propertyName: string): void;
 export function AfterBulkCreate(options: IHookOptions): Function;
 export function AfterBulkCreate(...args: any[]): void|Function {
-
-  if (args.length === 1) {
-
-    const options = args[0];
-
-    return (target: any, propertyName: string) =>
-      addHook(target, 'afterBulkCreate', propertyName, options);
-  } else {
-
-    const target = args[0];
-    const propertyName = args[1];
-
-    addHook(target, 'afterBulkCreate', propertyName);
-  }
+  return implementHookDecorator('afterBulkCreate', args);
 }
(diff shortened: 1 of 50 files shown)


RobinBuschmann:
Great job.

lib/annotations/hooks/AfterBulkCreate.ts
```diff
@@ -0,0 +1,21 @@
+import {IHookOptions} from "../../interfaces/IHookOptions";
+import {addHook} from "../../services/hooks";
+
+export function AfterBulkCreate(target: any, propertyName: string): void;
+export function AfterBulkCreate(options: IHookOptions): Function;
+export function AfterBulkCreate(...args: any[]): void|Function {
+
+  if (args.length === 1) {
```
RobinBuschmann: I would suggest to separate this (line 8-20) to an additional method in the hooks service cause these lines appear in every hook annotation. What do you think?

natesilva: How about this. No changes needed in the hooks service:

```typescript
export function AfterBulkCreate(target: any, propertyName: string): void;
export function AfterBulkCreate(options: IHookOptions): MethodDecorator;
export function AfterBulkCreate(...args: any[]): void|Function {

  return (target: any, propertyName: string) =>
      addHook(target, 'afterBulkCreate', propertyName, args.length ? args[0] : {});
}
```

The polymorphic function signature is confusing, so this keeps the logic related to that close to where the signature is defined instead of in a method somewhere else.

natesilva:
I messed up this branch. Re-opened the pull request as #101.

PR #100, opened 2017-08-16
support null value include in @Scopes 

hiradimir:
I got error of below log and fixed it.
This error occurred by null value in `@Scopes`.

```log
/home/user/app/node_modules/sequelize-typescript/lib/utils/object.js:10
            .getOwnPropertyNames(source)
             ^
TypeError: Cannot convert undefined or null to object
    at Function.getOwnPropertyNames (<anonymous>)
    at /home/user/app/node_modules/sequelize-typescript/lib/utils/object.js:10:14
    at Array.forEach (native)
    at deepAssign (/home/user/app/node_modules/sequelize-typescript/lib/utils/object.js:8:13)
    at assign (/home/user/app/node_modules/sequelize-typescript/lib/utils/object.js:34:21)
    at /home/user/app/node_modules/sequelize-typescript/lib/utils/object.js:11:46
    at Array.forEach (native)
```

### DIFF
--- a/lib/utils/object.ts
+++ b/lib/utils/object.ts
@@ -46,8 +46,10 @@ export function deepAssign(target: any, ...sources: any[]): any {
         } else if (sourceValue instanceof Date) {
 
           targetValue = new Date(sourceValue);
-        } else {
+        } else if (sourceValue === null) {
 
+          targetValue = null;
+        } else {
           deepAssign(targetValue, sourceValue);
         }
       } else {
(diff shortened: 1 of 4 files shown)


lib/utils/object.ts
```diff
@@ -46,8 +46,10 @@ export function deepAssign(target: any, ...sources: any[]): any {
         } else if (sourceValue instanceof Date) {
 
           targetValue = new Date(sourceValue);
-        } else {
+        } else if (sourceValue === null) {
```
hiradimir: above L39 condition of
`} else if (typeof sourceValue === 'object') {`
is true if sourceValue = null.

image

RobinBuschmann:
Thank you - looks good :)

PR #101, opened 2017-08-17
feature(hooks): add decorators for declaring hooks

### LINKED ISSUE #88: A `@Hook` decorator would be great

natesilva:
I had to re-do the pull request after I botched the original branch. The only change is that this centralizes the hook implementation logic into the hooks services file.

Fixes #88, replaces #97.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Decorators and some other extras for sequelize (v3 + v4).
    - [Multiple relations of same models](#multiple-relations-of-same-models)
  - [Model validation](#model-validation)
  - [Scopes](#scopes)
+ - [Hooks](#hooks)
  - [Why `() => Model`?](#user-content-why---model)
  - [Recommendations and limitations](#recommendations-and-limitations)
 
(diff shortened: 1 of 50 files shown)


natesilva:
Thank you!

RobinBuschmann:
Thank you for implementing this. It's a great feature!

RobinBuschmann:
I've made a beta release, in which this feature is included now :)

`npm install sequelize-typescript@0.5.0-beta.0`

mengkheang:
@RobinBuschmann can I use an async function like this?

```javascript
  @BeforeUpdate
  @BeforeCreate
  static async hashPassword(instance: User) {
    // this will be called when an instance is created or updated
    instance.password = await bcrypt.hash(instance.password, 10);
  }
```

natesilva:
@mengkheang Yes. A hook function may return a `Promise`, so `async` functions are fine. http://docs.sequelizejs.com/manual/tutorial/hooks.html#declaring-hooks

mengkheang:
@natesilva thank for the explaining :)

RobinBuschmann:
`sequelize-typescript@0.5.0` is released now 🎉

natesilva:
Thank you!

Issue #113, opened 2017-08-28
Compiler errors on 0.5.0-beta.0


bilby91:
Hello,

Today I tried to use the new beta release but I'm getting the following errors when compiling my project. 

```
$ tsc
node_modules/sequelize-typescript/lib/interfaces/IBaseIncludeOptions.d.ts(16,13): error TS2315: Type 'WhereOptions' is not generic.
node_modules/sequelize-typescript/lib/interfaces/IFindOptions.d.ts(9,13): error TS2315: Type 'WhereOptions' is not generic.
node_modules/sequelize-typescript/lib/interfaces/IFindOptions.d.ts(69,14): error TS2315: Type 'WhereOptions' is not generic.
node_modules/sequelize-typescript/lib/interfaces/IScopeFindOptions.d.ts(9,13): error TS2315: Type 'WhereOptions' is not generic.
node_modules/sequelize-typescript/lib/interfaces/IScopeFindOptions.d.ts(69,14): error TS2315: Type 'WhereOptions' is not generic.
```

From what I saw the IFindOptions is not a Generic type so I'm not sure what could be the problem. Previous version didn't have this issues on my project.

RobinBuschmann:
Hey @bilby91, which version of the sequelize typings are installed in your package? `sequelize-typescript@0.5.0-beta.0` requires `@types/sequelize@4.0.69`, in which `WhereOptions`is generic.

bilby91:
I just forced .69 version and the errors changed. Now I'm having only this error which makes sense:

```
$ tsc
node_modules/sequelize-typescript/lib/models/Model.d.ts(135,61): error TS2314: Generic type 'WhereOptions' requires 1 type argument(s).
error Command failed with exit code 2.
```

bilby91:
Also, when trying to run `tsc` from master branch I get this compiling errors:

```
➜  sequelize-typescript git:(master) tsc

43     options = Object.assign({}, optionsOrDataType);
       ~~~~~~~

lib/annotations/Column.ts(43,5): error TS2322: Type 'IPartialDefineAttributeColumnOptions | DataTypeAbstract' is not assignable to type 'IPartialDefineAttributeColumnOptions'.
  Type 'DataTypeAbstract' has no properties in common with type 'IPartialDefineAttributeColumnOptions'.


 25       (typeof config === "string") ?
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 26         config : // URI string
[...]

PR #115, opened 2017-08-29
Fix scope type

bilby91:


bilby91:
Should fix #113 . Please verify @RobinBuschmann if this is correct please!

RobinBuschmann:
Hey, thanks for fixing this. :)  Why do you add `<T extends Model<T>`? The generic isn't used in the signature of `scope`. Or do I overseen something?

bilby91:
@RobinBuschmann you are right. I originally thought that WhereOptions where of type `<T extends Model<T>`, then realized it's used as `any` on the other places. Will change that.

RobinBuschmann:
closes #113

bilby91:
@RobinBuschmann Can we do a beta.1 ?

RobinBuschmann:
It's done :)

PR #120, opened 2017-08-31
Fix constructor to support all Sequelize options

### LINKED ISSUE #119: Logging option not working in 0.5.0.beta.1
Hello,

The logging option it doesn't seem to be working correctly. This is the configuration I'm using to instantiate Sequelize.

```
{ 
      dialect: 'postgres',
      uri: 'postgres://suttna:development@localhost:5432/db_test',
      logging: false,
      modelPaths: [ '/Users/bilby91/Projects/app/bot/src/app/models' ],
      pool: { max: 5, min: 0, idle: 10000 } 
}
```

With that configuration I'm getting log messages every time an interaction with the db is happening.

bilby91:
Fixes #119

### DIFF
--- a/lib/models/v3/Sequelize.ts
+++ b/lib/models/v3/Sequelize.ts
@@ -11,27 +11,26 @@ import {ISequelizeAssociation} from "../../interfaces/ISequelizeAssociation";
 export class Sequelize extends SequelizeOrigin implements BaseSequelize {
 
   // to fix "$1" called with something that's not an instance of Sequelize.Model
-  Model: any = Function;
+  Model: any;
 
-  throughMap: { [through: string]: any } = {};
-  _: { [modelName: string]: typeof Model } = {};
+  throughMap: { [through: string]: any };
+  _: { [modelName: string]: typeof Model };
   init: (config: SequelizeConfig) => void;
   addModels: (models: Array<typeof Model> | string[]) => void;
   associateModels: (models: Array<typeof Model>) => void;
 
   constructor(config: SequelizeConfig | string) {
-
-    super(
-      (typeof config === "string") ?
-        config : // URI string
-        BaseSequelize.isISequelizeUriConfig(config) ?
-          config.uri : // URI string from ISequelizeUriConfig
-          BaseSequelize.prepareConfig(config) // Config object (ISequelizeConfig)
-    );
-
-    if (BaseSequelize.isISequelizeUriConfig(config)) {
-      this.options = {...this.options, ...config};
+    if (typeof config === "string") {
+      super(config)
+    } else if(BaseSequelize.isISequelizeUriConfig(config)) {
+      super(config.uri, config)
+    } else {
+      super(BaseSequelize.prepareConfig(config))
     }
+
+    this.throughMap = {}
+    this._ = {}
+    this.Model = Function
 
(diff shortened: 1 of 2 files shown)


bilby91:
@RobinBuschmann Hey, what do you think ? I think this is the last issue that is blocking us to get this in production! A beta.2 today would be AWESOME!

test/utils/sequelize.ts
```diff
@@ -19,6 +19,7 @@ export function createSequelize(useModelsInPath: boolean = true, define: DefineO
 export function createSequelizeValidationOnly(useModelsInPath: boolean = true): Sequelize {
 
   return new Sequelize({
+    dialect: 'sqlite',
```
RobinBuschmann: @bilby91 Whats the purpose of adding the `dialect`option here?

bilby91:
I'm not sure why it was working before. v4 requires a dialect to be explicitly defined.

RobinBuschmann:
Strange, there seem to have happened an issue earlier caused by another pull request I think.

There is a `ISequelizeValidationOnlyConfig` which is missing in:

`export type SequelizeConfig = ISequelizeConfig | ISequelizeUriConfig | ISequelizeDbNameConfig;`

When you add `ISequelizeValidationOnlyConfig` to `SequelizeConfig` it should work without `dialect: 'sqlite'`.

bilby91:
@RobinBuschmann Want me to add it ?

RobinBuschmann:
It depends... does it build without `dialect: 'sqlite',`? Cause this line does not fit the actual api, because it is `validationOnly` without the need to specify database configuration.

bilby91:
@RobinBuschmann Changing this `export type SequelizeConfig = ISequelizeConfig | ISequelizeUriConfig | ISequelizeDbNameConfig | ISequelizeValidationOnlyConfig;` didn't fix the issue. I think the problem is different. The thing is that now the url constructor passes the options always and in v4 you need to always define the dialect. Make sense ?

bilby91:
If we remove the `dialect: 'sqlite'` tests won't pass. And I think it makes sense that they don't pass without that option.

RobinBuschmann:
When `validationOnly: true` is defined, these options (e.g. `dialect`) will be added automatically: https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/models/BaseSequelize.ts#L61

RobinBuschmann:
`validationOnly: true` makes it possible to use `sequelize-typescript` without connecting to any database. Thats why it should not be necessary to define database options.

RobinBuschmann:
Here the reason why it compiles with `validationOnly: true` and without `dialect`: https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/models/Sequelize.d.ts#L11

bilby91:
@RobinBuschmann Good. Got it. It was not working localy because I though `npm run test` was compiling the code :(.

Now it should be good to go

bilby91:
Thanks for the awesome feedback loop @RobinBuschmann . Can we get a beta.2 🎉 ?

RobinBuschmann:
Thank you for fixing this :) 

Its done `sequelize-typescript@0.5.0-beta.2` 🎉

bilby91:
@RobinBuschmann Thanks!!

PR #126, opened 2017-09-04
When `IFindOptions.include = []` the `where` clause is ignored

izikl:
```
A = Model.findOne({
   where: { attr1: "1", attr2: "2" },
   include: []
});

B = Model.findOne({
   where: { attr1: "1", attr2: "2" },
   include: undefined
};
```

Expect: A = B
Actual: A ignores the where clause and B doesn't.


RobinBuschmann:
Hey @izikl thanks for contributing. Can you add tests, which demonstrates the issue?

izikl:
@RobinBuschmann added a test that is failing without the fix and works with it.

PR #148, opened 2017-09-26
tableName bug fix

skarahoda:
Setting tableName is unnecessary. It prevents to use underscoredAll option.

### DIFF
--- a/lib/annotations/Table.ts
+++ b/lib/annotations/Table.ts
@@ -20,8 +20,6 @@ export function Table(arg: any): void|Function {
 
 function annotate(target: any, options: IDefineOptions = {}): void {
 
-  if (!options.tableName) options.tableName = target.name;
-
   options.instanceMethods = target.prototype;
   options.classMethods = target;
 


RobinBuschmann:
Hey @skarahoda, in sequelize-typescript the class name should be set as default table name - it isn't unnecessary. Nevertheless I do understand that setting `tableName` in the table annotation prevents the usage of the underscoreAll option. Probably there is another way to achieve this

skarahoda:
Hey @RobinBuschmann thanks for response. For `sequelize.define` function  `tableName` is an optional parameter.  You can check the [link](http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-init) also. sequelize can set tableName with given options. If sequelize-typescript interferes this behavior, you are right. Please help me to fix this tiny problem so I can use `underscoredAll` option with seqelize-typescript in an elegant way.

RobinBuschmann:
@skarahoda I found a solution: Setting `freezeTableName` to true by default instead of using class name as default table name does the job and it should make it also possible to use `underscoredAll`.

I will create a pull request and close this one.

Thanks for bringing this up :)

skarahoda:
Thank you @RobinBuschmann  :)
Just for curiosity, when will you publish the new version?

RobinBuschmann:
I will try to publish a new beta release on this weekend.

PR #163, opened 2017-10-11
Fix tests due to IFindOptions change

bilby91:


bilby91:
@RobinBuschmann I think this will work. Really don't know why the previous build didn't catch the problem. 

I even spot a bug in a test with the IFindOptions change :).

PR #164, opened 2017-10-12
Fix Model#reload IFindOptions

bilby91:


bilby91:
Could it be a compiler version thing ? Not sure why it compiled correctly ...

bilby91:
Any chance we can get a new beta?

RobinBuschmann:
Done. Sry for the late response

Issue #157, opened 2017-10-07
Drop sequelize version 3 support for future releases


RobinBuschmann:
Supporting v3 and v4 at the same time makes the code base unnecessarily complex and improvements much more difficult. Since sequelize v4 is available for some month now, I think it makes sense to drop support for version 3.

@bilby91 @snewell92 @BruceHem What do you think about it?

bilby91:
I have the same opinion here! 

+1

Martín

On Sat, Oct 07, 2017 at 12:01 Robin Buschmann

<
mailto:Robin Buschmann <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

Supporting v3 and v4 at the same time makes the code base unnecessarily complex and improvements much more difficult. Since sequelize v4 is available for some month now, I think it makes sense to drop support for version 3.

https://github.com/bilby91 https://github.com/snewell92 https://github.com/brucehem
What do you think about it?

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/RobinBuschmann/sequelize-typescript/issues/157
, or

BruceHem:
@RobinBuschmann to be honest, I think it's best to drop support for v3.

nullpixel:
+1

snewell92:
I agree.

And if someone needs v3 support (lock in on production or something), it would be easy to either 1, pin `sequelize-typescript` to an older version with v3 support, or 2 - fork this repo, reset to a commit with v3 support, and just use that indefinitely (since sequelize v3 isn't going to undergo any major breaking changes, barring some security issue... in which case the sequelize team may still just advise to update to v4).

Let's do eeeeet!

RobinBuschmann:
Sequelize version 3 support will be dropped with `sequelize-typescript@1.0.0`. `sequelize-typescript@0.6.0` will be the last minor release before `1.0.0`.

RobinBuschmann:
Dropped on `sequelize-typescript@1.0.0-alpha-X`

PR #166, opened 2017-10-13
Loosen semver ranges on 3rd-party typings

### LINKED ISSUE #165: Bluebird version conflict
There appears to be a minor version incompatibility between the version of the Bluebird typings required by this project(3.5.4), and the latest version of those typings (3.5.15).  

 It may make sense to alter this project's `package.json` to play nicely with newer semver-minor releases of the Sequelize and Bluebird typings.  

`@types/sequelize` is _super-generous_, and actually just specifies `@types/bluebird: "*"` as a dependency). 

```
src/file.ts.ts(57,3): error TS90010: Type 'Bluebird<number>' is not assignable to type 'Bluebird<number>'. Two different types with this name exist, but they are unrelated.
  Types of property 'then' are incompatible.
    Type '{ <U1, U2>(onFulfill: (value: number) => U1 | Thenable<U1>, onReject: (error: any) => U2 | Thenab...' is not assignable to type '{ <U>(onFulfill?: ((value: number) => U | PromiseLike<U>) | undefined, onReject?: ((error: any) =...'.
      Type 'Bluebird<any>' is not assignable to type 'Bluebird<any>'. Two different types with this name exist, but they are unrelated.
        Types of property 'catch' are incompatible.
          Type '{ (onReject?: ((error: any) => any) | undefined): Bluebird<any>; <U>(onReject?: ((error: any) => ...' is not assignable to type '{ (onReject?: ((error: any) => any) | undefined): Bluebird<any>; <U>(onReject?: ((error: any) => ...'. Two different types with this name exist, but they are unrelated.
            Types of parameters 'onReject' and 'filter1' are incompatible.
              Type 'new (...args: any[]) => any' is not assignable to type '((error: any) => any) | undefined'.
                Type 'new (...args: any[]) => any' is not assignable to type '(error: any) => any'.
                  Type 'new (...args: any[]) => any' provides no match for the signature '(error: any): any'.
```

schmod:
Fixes #165, and possibly some other issues that may result from this project using a slightly older version of `@types/sequelize` and typescript.  

Due to one of these updates, the behavior of `getModels()` had to be slightly modified to conform with its type signature.  This should not be a breaking change for anybody.

Also fixed up a few tests that were using strange (and seemingly undocumented) variations of the `where:` operator.

### DIFF
--- a/lib/services/models.ts
+++ b/lib/services/models.ts
@@ -178,9 +178,14 @@ export function getSequelizeTypeByDesignType(target: any, propertyName: string):
  */
 export function getModels(arg: Array<typeof Model | string>): Array<typeof Model> {
 
-  if (arg && typeof arg[0] === 'string') {
+  if (arg) {
 
     return arg.reduce((models: any[], dir) => {
+
+      if (typeof dir !== 'string') {
+        models.push(dir);
+        return models;
+      }
 
       const _models = fs
         .readdirSync(dir as string)
--- a/package.json
+++ b/package.json
@@ -42,10 +42,10 @@
   "main": "index.js",
   "types": "index.d.ts",
   "dependencies": {
-    "@types/bluebird": "3.5.4",
-    "@types/node": "6.0.41",
-    "@types/reflect-metadata": "0.0.4",
-    "@types/sequelize": "4.0.73",
+    "@types/bluebird": "^3.5.15",
+    "@types/node": "^6.0.89",
+    "@types/reflect-metadata": "^0.0.4",
+    "@types/sequelize": "^4.0.76",
     "es6-shim": "0.35.3"
   },
   "devDependencies": {
@@ -74,9 +74,9 @@
     "sinon-chai": "2.8.0",
     "source-map-support": "0.4.14",
     "sqlite3": "3.1.8",
-    "ts-node": "3.0.4",
+    "ts-node": "3.3.0",
     "tslint": "4.3.1",
-    "typescript": "2.2.1",
+    "typescript": "2.5.3",
     "uuid-validate": "0.0.2"
   },
   "engines": {
--- a/test/specs/instance.spec.ts
+++ b/test/specs/instance.spec.ts
@@ -1014,10 +1014,10 @@ describe('instance', () => {
   });
 
   describe('complete', () => {
-
     it('gets triggered if an error occurs', () =>
+
       User
-        .findOne({where: ['asdasdasd']})
+        .findOne<User>({where: <any>{ bad: 'asdasdasd' }})
         .catch((err) => {
           expect(err).to.exist;
           expect(err.message).to.exist;
@@ -1936,9 +1936,9 @@ describe('instance', () => {
       ParanoidUser.create({username: 'cuss'})
         .then(() =>
           ParanoidUser.findAll<ParanoidUser>({
-            where: sequelize.and({
-              username: 'cuss'
-            })
+            where: {
+              [sequelize.Op ? sequelize.Op.and : '$and']: { username: 'cuss' }
+            }
           })
         )
         .then((users) => {
@@ -1947,9 +1947,9 @@ describe('instance', () => {
         })
         .then(() =>
           ParanoidUser.findAll({
-            where: sequelize.and({
-              username: 'cuss'
-            })
+            where: {
+              [sequelize.Op ? sequelize.Op.and : '$and']: { username: 'cuss' }
+            }
           })
         )
         .then((users) => {
@@ -1962,9 +1962,9 @@ describe('instance', () => {
       ParanoidUser.create({username: 'cuss'})
         .then(() =>
           ParanoidUser.findAll({
-            where: sequelize.or({
-              username: 'cuss'
-            })
+            where: {
+              [sequelize.Op ? sequelize.Op.or : '$or']: { username: 'cuss' }
+            }
           })
         )
         .then((users) => {
@@ -1973,9 +1973,9 @@ describe('instance', () => {
         })
         .then(() =>
           ParanoidUser.findAll({
-            where: sequelize.or({
-              username: 'cuss'
-            })
+            where: {
+              [sequelize.Op ? sequelize.Op.or : '$or']: { username: 'cuss' }
+            }
           })
         )
         .then((users) => {
--- a/lib/annotations/Column.ts
+++ b/lib/annotations/Column.ts
@@ -40,7 +40,7 @@ function annotate(target: any,
     };
   } else {
 
-    options = Object.assign({}, optionsOrDataType);
+    options = Object.assign({}, <IPartialDefineAttributeColumnOptions>optionsOrDataType);
 
     if (!options.type) {
       options.type = getSequelizeTypeByDesignType(target, propertyName);
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,11 +2,12 @@
   "name": "sequelize-typescript",
   "version": "0.6.0-beta.1",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@types/bluebird": {
-      "version": "3.5.4",
-      "resolved": "https://registry.npmjs.org/@types/bluebird/-/bluebird-3.5.4.tgz",
-      "integrity": "sha1-8SAWKwT9bVXhA0bX4ulu7UYrAs8="
+      "version": "3.5.15",
+      "resolved": "https://registry.npmjs.org/@types/bluebird/-/bluebird-3.5.15.tgz",
+      "integrity": "sha512-UA+JR/f3+gpEsmTHiJNN4YdhnYu1OgUTK0bF61qNR3X3/SKgod7pPkSlihAbdSGjUMpD3vb4pGd5EhVq6JvdIw=="
     },
     "@types/chai": {
       "version": "3.4.35",
@@ -18,13 +19,20 @@
       "version": "0.0.29",
       "resolved": "https://registry.npmjs.org/@types/chai-as-promised/-/chai-as-promised-0.0.29.tgz",
       "integrity": "sha1-Q9UokqqZjhhaPePiR37bhXO+HXc=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "@types/chai": "3.4.35",
+        "@types/promises-a-plus": "0.0.27"
+      }
     },
     "@types/chai-datetime": {
       "version": "0.0.30",
       "resolved": "https://registry.npmjs.org/@types/chai-datetime/-/chai-datetime-0.0.30.tgz",
       "integrity": "sha1-EhNEEGcsFkO9qlrKIzmw43F68Yg=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "@types/chai": "3.4.35"
+      }
     },
     "@types/lodash": {
       "version": "4.14.54",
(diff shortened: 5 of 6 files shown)


test/specs/instance.spec.ts
```diff
@@ -1014,10 +1014,10 @@ describe('instance', () => {
   });
 
   describe('complete', () => {
-
     it('gets triggered if an error occurs', () =>
+
       User
-        .findOne({where: ['asdasdasd']})
+        .findOne<User>({where: <any>{ bad: 'asdasdasd' }})
```
schmod: Incidentally, the newer Sequelize typings actually turn bad `where` clauses into a compiler error, making this sort of error a bit harder to trigger.

test/specs/instance.spec.ts
```diff
@@ -1936,9 +1936,9 @@ describe('instance', () => {
       ParanoidUser.create({username: 'cuss'})
         .then(() =>
           ParanoidUser.findAll<ParanoidUser>({
-            where: sequelize.and({
-              username: 'cuss'
-            })
+            where: {
+              [sequelize.Op ? sequelize.Op.and : '$and']: { username: 'cuss' }
```
schmod: Awkward workaround to simultaneously support Sequelize v3 and v4.  

`@types/sequelize` for v4 allows _some_ of the legacy operators, but not all of them (there might be a reason for it, but I can't find one).  I'll send them a separate PR to fix that.

snewell92: In #157 the project decided to drop v3 support - however since I don't think that has happened yet, I'm not sure what the best course of action for this code is. For now what you have is probably good, since the codebase currently tries to support both - but when we drop v3 we can get rid of the awkward workaround.

lib/services/models.ts
```diff
@@ -178,10 +178,15 @@ export function getSequelizeTypeByDesignType(target: any, propertyName: string):
  */
 export function getModels(arg: Array<typeof Model | string>): Array<typeof Model> {
 
-  if (arg && typeof arg[0] === 'string') {
+  if (arg) {
 
     return arg.reduce((models: any[], dir) => {
 
+      if (typeof dir !== 'string') {
```
snewell92: @RobinBuschmann do we have tests on this method? (I don't think I saw any in `model.spec.ts`?) This would be good to have some tests. (array with Model and String elems / only string elems / only Model elems / empty array / null / object).

snewell92:
@schmod Glad to have a new first time PR 😺I think it's golden, just not sure if we could add tests to increase coverage or not (hence ping to Robin), but Codecov has been known to be finicky about small things ¯\\\_(ツ)_/¯

@RobinBuschmann could you look over this when you can, just to look over the change in `getModels` - then I believe this can be merged.

test/specs/instance.spec.ts
schmod: Yep, but as long as we've got tests targeting v3....

package.json
```diff
@@ -42,10 +42,10 @@
   "main": "index.js",
   "types": "index.d.ts",
   "dependencies": {
-    "@types/bluebird": "3.5.4",
-    "@types/node": "6.0.41",
-    "@types/reflect-metadata": "0.0.4",
-    "@types/sequelize": "4.0.73",
+    "@types/bluebird": "^3.5.15",
```
RobinBuschmann: Let us keep the exact versions here. Here is a discussion about it: https://github.com/RobinBuschmann/sequelize-typescript/issues/142

lib/services/models.ts
RobinBuschmann: @snewell92 No, we don't have any for this method only. @schmod  Can you add some tests as a part of this PR? This would also increase coverage :)

RobinBuschmann:
Maybe we can get rid of setting the bluebird typings explicitly as a dependency, cause they are required by the sequelize typings anyway. What do you think, @snewell92  @schmod ?

snewell92:
If we drop bluebird (and remove its use from `Model.d.ts` and `BaseModel.d.ts`), will that allows users to use bluebird in their applications and the promises that sequelize-typescript / sequelize returns will be promises? If so, that sounds great! I'm always down for removing code and dependencies if the library/app can still do the same thing without it.

schmod:
Sequelize already uses Bluebird promises "out of the box", and the typings accurately reflect this.  You should theoretically be able to do this already.

Unfortunately, there have been "semver-minor" changes to Bluebird's typings that appear to be incompatible with each other, which is why this does not currently appear to work in many cases.

---

Getting rid of the Bluebird typings probably won't be practical, unless we can explicitly import them from `@types/sequelize` (which feels a bit hacky).  

If we're maintaining a hard lock on a specific version of `@types/sequelize`, we might as well just make the version specifier of `@types/bluebird` in our `package.json` match theirs.  

---

Given the recurrence of these issues, I think our docs need to explain this project's relationship with `@types/sequelize`, and advise users to `import { Sequelize } from 'sequelize-typescript` instead of loading the upstream sequelize types directly (given the likelihood that they'll import a different version).  We may be able to enforce this via a `peerDependency` or `optionalDependency`.

snewell92:
> we might as well just make the version specifier of @types/bluebird in our package.json match theirs

I think it makes sense to do what `@types/sequelize` is doing with regards to typing versions, especially since we depend on a specific version of it (due to incompatibility issues).

> We may be able to enforce this via a peerDependency

*A peer dependency!* That sounds very reasonable. Now that I think about it... yes. That is exactly the relationship sequelize-typescript expects with sequelize (since it doesn't have a dep on sequelize, but won't work without it).

RobinBuschmann:
Regarding bluebird: Hmm, I probably miss the point here or I was not clear enough. The bluebird typings will be installed anyway, cause they are a dependency of the sequelize typings. So there is nothing left to do... or maybe I don't understand what are you going to say? 😱

devuxer:
@schmod, until this is resolved, could you tell me what version of sequelize to downgrade to so that the types in sequelize-typescript will be compatible with the types in sequelize?

schmod:
> The bluebird typings will be installed anyway

I don't think this is a great assumption to make.  It might not always be true, and contradicts the conventions typically used by Node.JS projects -- I can't think of any projects that `require()` a 3rd-party module without listing it in their `package.json` file.  

For example, this is a perfectly valid tree that NPM (or some other package manager) could produce:

```
.
└── node_modules
    ├── bluebird
    ├── @types
    │   └── sequelize
    │       └── node_modules
    │           └── @types/bluebird
    └── sequelize-typescript
```

In this example, the [`require.resolve()`](https://nodejs.org/api/modules.html#modules_all_together) algorithm would be unable to locate `@types/bluebird` from the root of `sequelize-typescript`.

RobinBuschmann:
I can think of two scenarios, where this would happen: on an old node version or if the bluebird typings are already installed. Is there another one ?

Anyway, you're absolutely right, an explicit solution is better.

However I think we can omit the bluebird typings. Because the sequelize typings exports them. So that sequelize-typescript will use promises like this `import {Promise} from 'sequelize'; Do I miss something?

snewell92:
Just revisiting - I still really like @schmod 's suggestion of a peer dep to sequelize, and matching bluebird typings to what [@types/sequelize lists](https://github.com/types/sequelize/blob/master/package.json#L8), and of course matching bluebird to what [sequelize lists](https://github.com/sequelize/sequelize/blob/master/package.json#L39) too, so that when we bump a sequelize version, we then must match bluebird and its typings for the appropriate version of sequelize we're bumping up too.

This will ensure correctness in our types.

@RobinBuschmann I think your two scenarios fail to capture the complexity of npm's package system. There are a lot of different scenarios:

1. No bluebird installed at the root
2. bluebird installed at the root
  a. bluebird at root matches our version
  b. bluebird at root does not match

Then consider, in the dependency tree, since sequelize installs bluebird too, if our bluebird and its types match sequelize's, and if either of those (sequelize or our's) match the root import. Not to mention if there are other packages that may or may not wrap, call into, or use sequelize/sequelize-typescript (like `feathers-sequelize` for example)

Thus we reach a combinatorial explosion of cases. (I think)

The typescript compiler can get the appropriate types at each level, but the problem is that if bluebird introduces some breaking change, and if sequelize or sequelize-typescript call into their own copy of bluebird down the dep tree - then typescript might report something that doesn't actually exist (since at the root, bluebird is available), or an API has changed and won't work as expected. I'm not actually sure how this would work, and am interested in a deeper case study, but I don't think that's necessary to get to a conclusion here.

Now, I can see how just dropping bluebird might seem to sidestep the issue, but then it's just another subtle problem, because we wrap sequelize and our code base calls into sequelize, we wouldn't be exposing the actual real type of the promises that sequelize returns, and if we ever create a promise to do work, it could (would it have to be?) be native instead of bluebird, which would mix promise types. Again, not 100% on all that, but just thinking about it makes me squirm.

@schmod if I've mistakenly analyzed anything, please rebuketh me! #learning #discussion 

=> keep bluebird, match its version to sequelize

RobinBuschmann:
Ok, lets define an explicit dependency of @types/bluebird. The question is, do we need to set it as a peer dependency or not? To ensure that the @types/bluebird version matches the version required by sequelize, we need to use the exact same version, right? So if this is needed anyway, is there a benefit from using it as peer dependency?

snewell92:
Good question. If we list bluebird as a peer dep, we force the user to use bluebird. Since `sequelize` doesn't do that, I don't think we should do that. We should report correct types from sequelize, since they use bluebird, but I think it would be fine if users use native promises, then they would just lose out on bluebird features but still be able to use `Promise.all`, `p.then`, `Promise.resolve` and friends, as I think bluebird was designed with that in mind (progressive library, legacy user).

Just an example package.json so I can wrap my head around it more:

**package.json** snippet
```json
  "dependencies": {
    "@types/sequelize": "^4.0.76",
    "bluebird": "^3.4.6",
    "@types/bluebird": "^3.5.5"
  },
  "peerDependencies": {
    "sequelize": "4.22.6"
  },
```

@schmod @RobinBuschmann  perhaps our `sequelize` peer dep should just be a minimum range? like `4.22.x`? Idk how strict we want to be. Pinning sequelize to a specific version makes it easier for us, but less flexible for the end user. I like having `@types` be flexible so we don't have to bump up anytime someone lands a PR into sequelize typings, but sequelize might be a little hairier... perhaps we can rely on `sequelize` following semver, and just take patches?

schmod:
I honestly don't see a problem being somewhat permissive with the version ranges of our `sequelize` dep (whether it's a peer dep or regular dep).  I think something like `"~4.22.0"` would be totally acceptable.

`@types/sequelize`, on the other hand, has historically caused us a lot of problems, because their "minor" releases have sometimes made major updates to the typings, which is obviously risky for us, given our extremely close and complicated relationship with those types.

snewell92:
Ah, okay. So something like this:

**package.json** snippet:
```json
  "dependencies": {
    "bluebird": "3.4.6",
    "@types/bluebird": "3.5.15",
    "@types/sequelize": "4.0.76"
  },
  "peerDependencies": {
    "sequelize": "~4.22.0"
  }
```

What do y'all think? 👍 for ya, 😕 for wat

RobinBuschmann:
Just this week I've noticed this tsc feature: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

So my question: Should we use this and put the type dependencies (`@types/bluebird`, `@types/sequelize`) in our build to minimize dependencies of this package. What do you think? @schmod @snewell92

snewell92:
I sort have used `typeRoots`, but not in a library context.

I'd be interested to see how typescript handles several cases, I might have time to play with this kind of stuff next week. I would try by specifying the type roots, and then specifying no type roots, and having a consuming project both match and then not match the bluebird / sequelize typings - just to see what happens.

I don't think the use of `typeroots` will solve the problem though.

RobinBuschmann:
> I'd be interested to see how typescript handles several cases, I might have time to play with this kind of stuff next week. I would try by specifying the type roots, and then specifying no type roots, and having a consuming project both match and then not match the bluebird / sequelize typings - just to see what happens.

This would be nice :)

> I don't think the use of typeroots will solve the problem though.

Wasn't intended to be a fix for this issue. Probably not the right place to mention this feature.

RobinBuschmann:
With version `1.0.0-alpha.8` sequelize-typescript uses the typings bundled with sequelize. So the `@types/sequelize` and `@types/bluebird` dependencies are removed there.

Issue #177, opened 2017-10-24
Glob based model paths


onebesky:
Some projects keep models within modules right next to controllers. Current implementation of sequelize-typescript allows to specify only path to folder with models.

What do you think about adding support for the following configuration:
```
modelPaths: [__dirname + '/**/*.model.ts']
```

RobinBuschmann:
Sounds good. I think we should implement this next.

RobinBuschmann:
#24

victorivens05:
Hey guys. 
I made some simple changes and was about to make sure they worked, before I opened a PR, when I found this discussion. 
So, what will be done? Add the include option in the sequelize constructor or implement glob in the `modelPaths`?

RobinBuschmann:
Using `modelPaths` for an additional globing functionality should be fine. So that using a path to a directory with `modelPaths` should continue working as well. 

@victorivens05 What is you suggestion?

victorivens05:
@RobinBuschmann in my tests I managed to use glob and folder together.
Using glob, if you pass a folder: `c:\project\models` it doesn't return anything, but there's a method called `glob.hasMagic` with which we can detect if the user passed a folder or a glob.
If the user passed a folder, we concatenate with `\*` and get every file in the folder.

It should work, unless some user has a folder called `!(folder)`, `@(folder)` or `[folder]`

The problem right now is that I work alone, so I don't have much experience working with teams. The tests seems to not be working and I don't know if I should create a branch and submit a PR.

RobinBuschmann:
@victorivens05 Sounds good! Don't worry. Create your PR and we will have a look :) If there is something that need to be adjusted we can discuss that in the context of your PR. If you have any questions don't hesitate to ask.

victorivens05:
I just submitted the PR. I already think some stuff should be done differently. I didn't create a test, I'd like to create one, but I would need some help to figure out where and how it should be done.

snewell92:
Worth mentioning from [this other issue thread](https://github.com/RobinBuschmann/sequelize-typescript/issues/24#issuecomment-308157677)

> I think it would be much clearer to use only paths for modelPaths option and no globs. Using globs in include, exclude with an absolute path should work.

It might be a good idea to leave `modelPaths` alone, and then add `include` and `exclude` options that do exactly what your PR is doing for `modelPaths`.

I'll defer to @RobinBuschmann - it will be super nice to have globs either which way imo. :D

[...]

Issue #24, opened 2017-06-13
Add excludes / includes option


snewell92:
Based on the discussion within #23 we can add a feature in the `ISequelizeOptions` for support of `includes/excludes` options. To fully support this, we should add support for [globs](https://www.npmjs.com/package/glob).

Still a bit of discussion as to whether or not `modelPaths` should only be paths - since it would be easier to implement if we could just concatenate both the `includes` array and `modelPaths` array together and have logic in the reduce loop to discriminate between directories, files, or a glob pattern. Or possibly use the reduce loop to pass through the array once to expand all glob patterns/directories, building an array of only files, then after the reduce loop map over them with `require` - although this would require first pulling out all passed in `Models` from the array so we just get an array of models and an array of strings. Meh, maybe not worth the effort.

The excludes option seems far easier to grapple with. We already have a filter in `models.ts -> getModels` function so we just add that in.

RobinBuschmann:
Yeah, I think it would be much clearer to use only paths for `modelPaths` option and no globs. Using globs in `include`, `exclude` with an absolute path should work.

snewell92:
:+1: Sound great.

RobinBuschmann:
I think this can be closed, since globs are solving this mostly.

PR #181, opened 2017-10-30
Add the hability to parse Glob and Folder path in modelPaths

victorivens05:
Changed the usage of FS to GLOB. If the Path isn't a glob, it's a folder
path and to work with glob, must be turned into a glob (by adding /* in
the end).
I think the check if it `hasMagic` should be an outside function,
chained in a `map` before `arg.reduce`.

### DIFF
--- a/lib/services/models.ts
+++ b/lib/services/models.ts
@@ -1,5 +1,5 @@
 import 'reflect-metadata';
-import * as fs from 'fs';
+import * as glob from 'glob';
 import * as path from 'path';
 import {DataTypeAbstract, DefineOptions} from 'sequelize';
 import {Model} from "../models/Model";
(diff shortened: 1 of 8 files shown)


victorivens05:
#177

RobinBuschmann:
@victorivens05 You could setup testing as follow:
- Create a sub directory in `/test/models/` - for example `/test/models/globs`
- Create additional sub directories and choose appropriate names matching your test cases - for example `/test/models/globs/match-dir-only` or `/test/models/globs/match-sub-dir-files`
- Put your specs into `/test/specs/models/sequelize.spec.ts` -> Check if the models are loaded properly :)

victorivens05:
@RobinBuschmann I created some tests, I'm not sure if they're enough. Now how do I add them to the PR? Or I have to create another?

RobinBuschmann:
@victorivens05 Pushing you're changes to the remote branch of the PR should be enough :) Since you've created a PR from `victorivens05:master`, pushing to this branch should work

RobinBuschmann:
@victorivens05 Looks good 👍 I will merge it  🎉

RobinBuschmann:
closes #177 and closes #24

PR #189, opened 2017-11-03
Type inference for static Model methods

schmod:
Adds type inference to most static methods on `Model` and its subclasses (`create()`, `findAll()`, etc) that previously required user-supplied generic parameters to correctly yield the correct return type.

`Person.findAll<Person>` simply becomes `Person.findAll()`, and TypeScript will automatically infer the correct return type.  The old type signature will also continue to work.  See https://github.com/Microsoft/TypeScript/issues/5863 for an explanation of what's going on under the hood that allows this to work.

New overloads have been added to `Model.create<T extends Model<T>, A>(vals: A) : T` and its cousins, so that `Model.create<A>(vals: A) : T` may also be used in as a shorthand for when developers want to provide strict typings for `A`, but don't need to override the automatic inference of `T`.

Updates tests to run against the current version of Sequelize v4, and uses newer Bluebird type definitions (a milder version of #166 until we can figure out what to do there -- necessary because the older version had an inaccurate definition of `Promise.all`).

### DIFF
--- a/lib/models/Model.d.ts
+++ b/lib/models/Model.d.ts
@@ -303,7 +303,7 @@ export declare abstract class Model<T> extends Hooks {
   /**
    * Builds a new model instance and calls save on it.
    */
-  static create<T extends Model<T>>(this: (new () => T), values?: T, options?: ICreateOptions): Promise<T>;
+  static create<T extends Model<T>>(this: (new () => T), values?: any, options?: ICreateOptions): Promise<T>;
   static create<T extends Model<T>, A>(this: (new () => T), values?: A, options?: ICreateOptions): Promise<T>;
   static create<A>(this: (new () => T), values?: A, options?: ICreateOptions): Promise<T>;
 
--- a/lib/models/Model.d.ts
+++ b/lib/models/Model.d.ts
@@ -57,7 +57,7 @@ export declare abstract class Model<T> extends Hooks {
    * @param schema The name of the schema
    * @param options
    */
-  static schema<T extends Model<T>>(schema: string, options?: SchemaOptions): Model<T>;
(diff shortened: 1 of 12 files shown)


lib/models/Model.d.ts
```diff
@@ -289,30 +289,35 @@ export declare abstract class Model<T> extends Hooks {
...
+  static bulkBuild<T extends Model<T>>(this: (new () => T), records: any[], options?: IBuildOptions): T[];
+  static bulkBuild<T extends Model<T>, A>(this: (new () => T), records: A[], options?: IBuildOptions): T[];
+  static bulkBuild<A>(this: (new () => T), records: A[], options?: IBuildOptions): T[];
 
   /**
    * Builds a new model instance and calls save on it.
    */
-  static create<T extends Model<T>>(values?: any, options?: ICreateOptions): Promise<T>;
-  static create<T extends Model<T>, A>(values?: A, options?: ICreateOptions): Promise<T>;
+  static create<T extends Model<T>>(this: (new () => T), values?: T, options?: ICreateOptions): Promise<T>;
```
RobinBuschmann: It should be `values?: any` instead of `values?: T`

RobinBuschmann:
Hey @schmod really really good feature. Didn't know that this is possible. Thank you for implementing this. I only found one issue in the typings: 
https://github.com/RobinBuschmann/sequelize-typescript/pull/189/files#diff-1c485f95a55d5ce0466f63a966ced297R306

lib/models/Model.d.ts
schmod: Indeed.  I'm not sure how that got flipped!  Will revise.

RobinBuschmann:
`0.6.0-beta.3` released

PR #191, opened 2017-11-07
fix build errors

### LINKED ISSUE #190: Typescript compiler output bunch of errors for version 0.6.0-beta.3
Anybody familiar with this error when compiling a project running `sequelize-typescript v 0.6.0-beta.3`:

```
node_modules/sequelize-typescript/lib/models/Model.d.ts(133,129): error TS2693: 'T' only refers to a type, but is being used as a value here.

133   static scope<T extends Model<T>>(this: (new () => T), options?: string | string[] | ScopeOptions | WhereOptions<any>): typeof T;
```

Plus many others similar to this one.

Is there any settings or configuration setup that I am missing?

schmod:
Fixes #190

Adds a documentation note about the fact that scopes effectively break typesafety unless an explicit cast is provided.

### DIFF
--- a/test/tsconfig.json
+++ b/test/tsconfig.json
@@ -6,10 +6,9 @@
     "emitDecoratorMetadata": true,
     "sourceMap": true,
     "strictNullChecks": false,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
     "pretty": true,
-    "skipLibCheck": true,
-    "lib": ["es2015"]
+    "lib": ["es2017"]
   },
   "include": [
     "./**/*"
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,8 +9,7 @@
     "strictNullChecks": true,
     "noUnusedLocals": true,
     "pretty": true,
-    "skipLibCheck": true,
-    "lib": ["es2015"]
+    "lib": ["es2017"]
   },
   "include": [
     "lib/**/*"
--- a/README.md
+++ b/README.md
@@ -553,6 +553,14 @@ export class ShoeWithScopes extends Model<ShoeWithScopes> {
 }
 ```
 
+Note that `Model.scope()` and `Model.unscoped()` are **not** typesafe.  Their return values must be explicily cast back to the correct Model type:
+
+```typescript
+(ShoeWithScopes.scope('full') as typeof ShoeWithScopes)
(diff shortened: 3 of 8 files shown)


tsconfig.json
```diff
@@ -9,8 +9,7 @@
     "strictNullChecks": true,
     "noUnusedLocals": true,
     "pretty": true,
-    "skipLibCheck": true,
```
RobinBuschmann: @schmod Is there a reason why you removed this line?

test/tsconfig.json
```diff
@@ -6,10 +6,9 @@
     "emitDecoratorMetadata": true,
     "sourceMap": true,
     "strictNullChecks": false,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
     "pretty": true,
-    "skipLibCheck": true,
```
RobinBuschmann: @schmod Same here, see below

RobinBuschmann:
@schmod Thanks for fixing it so fast :)

tsconfig.json
schmod: When set to `true`, tsc doesn't validate/check our `.d.ts` files.  This is how my original bug slipped through.

Because we're writing a bunch of our own `.d.ts` files and are semver-strict with most of our dependencies, it seems like there are very few benefits to skipping these checks.

test/tsconfig.json
```diff
@@ -6,10 +6,9 @@
     "emitDecoratorMetadata": true,
     "sourceMap": true,
     "strictNullChecks": false,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
schmod: Also, this was needed to get TS 2.6.x to not complain about some of our test classes that have decorators on them, but are otherwise "unused."

tsconfig.json
RobinBuschmann: Ah ok. Not what I've expected. I assumed that only libraries aren't checked with `skipLibCheck: true` as the name stated.

That explains why other typing related issues weren't revealed by running `tsc`.

Thanks for clearing.

test/tsconfig.json
RobinBuschmann: Can you explain this a little bit more or provide a link? Is this a bug since 2.6?

PR #194, opened 2017-11-08
$set should accept an array of ids

natesilva:
The `$set` method should accept an array of ids (`number[]`) as its `instances` argument. Currently it only accepts a single instance or an array of instances.

Passing in an array of primary keys instead of models is documented here: http://docs.sequelizejs.com/class/lib/associations/belongs-to-many.js~BelongsToMany.html#instance-method-set

### DIFF
--- a/test/specs/instance.spec.ts
+++ b/test/specs/instance.spec.ts
@@ -10,7 +10,7 @@ import {
 import {User} from "../models/User";
 import {TimeStampsUser} from "../models/TimeStampsUser";
 import {Book} from "../models/Book";
-import {Page} from "../models/Page";
+import { Page } from '../models/Page';
 import {Team} from "../models/Team";
 import {Player} from "../models/Player";
 import {Shoe} from "../models/Shoe";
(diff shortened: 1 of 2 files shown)


natesilva:
Note, I can’t currently run the unit tests. Running against the current `master` I get:

```
test/specs/instance.spec.ts(26,13): error TS2529: Duplicate identifier 'Promise'. Compiler reserves name 'Promise' in top level scope of a module containing async functions.

26 import * as Promise from 'bluebird';
```

RobinBuschmann:
I couldn't reproduce the error logged in the travis build, but I will further try. Can you reproduce it locally when running `tsc --project test --target es6`?

natesilva:
How strange. I’m no longer seeing the error. It worked doing `tsc --project test --target es6` and doing `npm install && npm run build && npm run cover`.

Can you trigger the Travis build again?

test/specs/instance.spec.ts
```diff
@@ -684,6 +684,27 @@ describe('instance', () => {
         });
     });
 
+    it('should udpate associations when given ids instead of models', async() =>
```
RobinBuschmann: Ah, I think it is the usage of `async` here, which implicitly requires the native `Promise` api and conflicts with the bluebird implementation. Removing this, should fix it.

Sry for the late response

snewell92:
@natesilva As per [robin's comment](https://github.com/RobinBuschmann/sequelize-typescript/pull/194#discussion_r151515184) if you remove `async` the test should pass - LGTM once that is gouda 👍

natesilva:
👍 I’ll do that, probably tomorrow.

RobinBuschmann:

PR #201, opened 2017-11-20
add static associations property to Model typing (closes #199)

### LINKED ISSUE #199: Typesafe way of getting Associations
I'm looking at **graphql-sequelize** and trying to get things working with **graphql-typescript** here: https://github.com/mickhansen/graphql-sequelize/blob/master/docs/relay.md.

The example shows passing in an association like `User.Tasks` so the library can automatically fetch them. After reading through the documentation and debugging/profiling I've found thats a **HasMany** association object and was trying to achieve the same thing here.

I've got this:
```ts
@Table
export class Seed extends Model<Seed> {
  @PrimaryKey
  @AutoIncrement
  @Column
  id: number;
 
  @Column
  name: string;

  @BelongsTo(() => User)
  user: User;
}

@Table
export class User extends Model<User> {
  @PrimaryKey
  @AutoIncrement
  @Column
  id: number;

  @HasMany(() => Task)
  tasks: Task[];
}

...

sequelize.addModels([ User, Task ]);
```

It seems like a property called `associations` magically appears on my class after calling addModels but it's not safe to get without casting to `any` first like:
```ts
import { User } from '../models/User';

console.log((<any>User).associations.tasks);
```

Any ideas/thoughts on how I can PR an improvement for this?

intellix:


RobinBuschmann:
@intellix This will be in the next version of `sequelize-typescript`

PR #202, opened 2017-11-20
Add npm5 git support

intellix:
Wanted to use my update but you haven't released it yet so would be nice to just install it directly from your git without the need to publish (something I miss from bower) :)

You can install the latest like:
```
npm i github:intellix/sequelize-typescript#npm5-git-support
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "exec-tests": "mocha",
     "test": "npm run test-v4 && npm run test-v3",
     "cover": "npm run cover-v4 && npm run cover-v3",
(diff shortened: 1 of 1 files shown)


bilby91:
Hey @RobinBuschmann I think this is a better idea than the automatic release that we have been discussing. 

npm docs state:

```
Git dependencies with prepare scripts will now have their devDependencies installed, and their prepare script executed as if under npm pack.
```

We would be only affecting users that install via git. I like this idea @intellix !

RobinBuschmann:
Looks good! Let's do it!

PR #206, opened 2017-11-26
closes #204 scope typings

### LINKED ISSUE #204: Since #189 scope typings only work together with as
`Movie.scope(req.query['scope'] || 'defaultScope').findOne()` throws
`(Movie.scope(req.query['scope'] || 'defaultScope') as typeof Movie).findOne(...)`

Improve typings!

RobinBuschmann:
Hey @schmod, I've found a better solution regarding the scope typings issue. Do you see any issue with this? Thanks for helping


schmod:
Oh, wow.  That's clever. 

I had no idea something like that would work.  Nice!!!

PR #227, opened 2017-12-11
Fix IScopeFindOptions typings

### LINKED ISSUE #226: Order DESC in scope not possible with Sequelize.col
What I want to do is get the average value of a column grouped by an id. This is like a rating. Therefore I would need to get this code-snippet to work:
 ```javascript
@Scopes({
    ranked: {
        attributes: [[Sequelize.fn('AVG', Sequelize.col('review.points')), 'avg']],
        order: [[Sequelize.col('avg'), 'DESC']],
    group: 'myId'
    }
})
```

This right now doesn't work but gives me the error:

```
[ts]
Argument of type '{ ranked: { attributes: [fn, string][]; order: (string | col)[][]; group: string; }; }' is not assignable to parameter of type 'IDefineScopeOptions'.
  Property 'ranked' is incompatible with index signature.
    Type '{ attributes: [fn, string][]; order: (string | col)[][]; group: string; }' is not assignable to type 'Function | IScopeFindOptions | undefined'.
      Type '{ attributes: [fn, string][]; order: (string | col)[][]; group: string; }' is not assignable to type 'IScopeFindOptions'.
        Types of property 'order' are incompatible.
          Type '(string | col)[][]' is not assignable to type 'string | col | literal | (string | number | ModelClassGetter | { model: ModelClassGetter; as?: st...'.
            Type '(string | col)[][]' is not assignable to type '(string | col | literal | (string | number | ModelClassGetter | { model: ModelClassGetter; as?: s...'.
```

I think this is due to wrong typings, because when I build it, everything works as expected. So maybe in [IScopeFindOptions](https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/interfaces/IScopeFindOptions.ts) the following has to be changed:

```typescript
order?: string | col | literal | Array<string | number | ModelClassGetter | { model: ModelClassGetter, as?: string }> |
  Array<string | col | literal | Array<string | number | ModelClassGetter | { model: ModelClassGetter, as?: string }>>;
```

to ->

```typescript
order?: string | col | literal | Array<string | col | number | ModelClassGetter | { model: ModelClassGetter, as?: string }> |
  Array<string | col | literal | Array<string | col | number | ModelClassGetter | { model: ModelClassGetter, as?: string }>>;
```

Any thoughts on this? I can add an according mergerequest. I'm kinda new to sequelize in general so maybe I'm also missing something and it's possible to solve my problem easier.

patsimm:
Fixes #226

### DIFF
--- a/lib/interfaces/IScopeFindOptions.ts
+++ b/lib/interfaces/IScopeFindOptions.ts
@@ -41,8 +41,8 @@ export interface IScopeFindOptions extends LoggingOptions, SearchPathOptions {
    * first element is the column / function to order by, the second is the direction. For example:
    * `order: [['name', 'DESC']]`. In this way the column will be escaped, but the direction will not.
    */
-  order?: string | col | literal | Array<string | number | ModelClassGetter | { model: ModelClassGetter, as?: string }> |
-    Array<string | col | literal | Array<string | number | ModelClassGetter | { model: ModelClassGetter, as?: string }>>;
+  order?: string | col | literal | Array<string | col | number | ModelClassGetter | { model: ModelClassGetter, as?: string }> |
+    Array<string | col | literal | Array<string | col | number | ModelClassGetter | { model: ModelClassGetter, as?: string }>>;
 
   /**
    * Limit the results


RobinBuschmann:
@patsimm, thanks for fixing. I will look into the build issue and will come back to you soon.

PR #235, opened 2017-12-14
Remove scope typing note

Chocobozzz:
Not necessary after https://github.com/RobinBuschmann/sequelize-typescript/pull/206/files


RobinBuschmann:
Thanks for adjusting this. Unfortunately this is blocked by a build error #236 . I need to look into this soon.

Issue #239, opened 2017-12-15
Improve any types for values/records in Models.d.ts 


Danlock:
The types for functions that have to do with creation such as Model.create or Model.build have a any type on their values: parameters. Is there a reason why that can't be Partial<T>  to get some compile time verification of model attributes on such methods?

I know that this includes function names as well, but it's better than any.
I'm going to do this on my project anyway, so let me know if you would accept a PR for this.

RobinBuschmann:
Hey @Danlock. You mean this https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/models/Model.d.ts#L308 (`values?: any`) for example, right? You would recommend to change this to `values?: Partial<T>`? Yeah, sounds good. @bilby91 Do you see any issues with this=

BruceHem:
@RobinBuschmann currently i use Partial<Pick<T, 'id' | 'etc'>>

Danlock:
So every time you use a method like create, you have a big list of all the attributes right next to the function call? That sounds cumbersome, but I guess it's more typesafe.

Is it possible to select against subset of Model keys that are always there so that the remainder are 'legit' columns? Something like a Difference of keys between whatever Model is being passed in vs the standard Model functions and attributes we know is always there.
I'm meaning something like 
```
type unwantedAttributes = '$get' | '$add' | 'beforeCreate' | ...
Partial<Difference<T,unwantedAttributes>> 

```

If this is not possible, Partial<T> might be the only practical solution without any knowledge of the user's attributes.

Danlock:
Ok, so I looked around for type differences and stumbled upon https://github.com/Microsoft/TypeScript/issues/18567.

With that I think I found the closest solution to getting the attributes for typesafety without needing to manually specify database columns.

```
type Diff<T extends string, U extends string> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
type Omit<T, K extends keyof T> = { [P in Diff<keyof T, K>]: T[P] };

type GetAttributes<T extends Model<any>> = Omit<
  T,
  | '$add'
  | '$set'
  | '$get'
  | '$count'
  | '$create'
  | '$has'
  | '$remove'
  | 'isNewRecord'
  | 'Model'
  | 'sequelize'
  | 'where'
  | 'getDataValue'
  | 'setDataValue'
  | 'get'
  | 'set'
  | 'setAttributes'
  | 'changed'
  | 'previous'
  | 'save'
[...]

PR #249, opened 2017-12-21
narrow down types for attributes passed to functions like Model.create/Model.build etc

Danlock:
for issue https://github.com/RobinBuschmann/sequelize-typescript/issues/239
I wanted to typecheck upsert values as well, but it doesn't seem to have a generic overloaded method like all the others.

### DIFF
--- a/lib/models/Model.d.ts
+++ b/lib/models/Model.d.ts
@@ -24,6 +24,18 @@ import {ICountOptions} from '../interfaces/ICountOptions';
 type ForTypeOf<T> = {[P in keyof T]: T[P]};
 type NonAbstractTypeOfModel<T> = (new () => T) & ForTypeOf<typeof Model>;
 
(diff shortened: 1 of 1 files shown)


RobinBuschmann:
@Danlock Thanks for implementing :) I will look at it very soon!

RobinBuschmann:
@Danlock Thanks again for your implementation. Looks good to me. There is one question left from my side: Why do you name the type `GetAttributes`? Why not `ModelAttributes` or something like that? :)

Danlock:
Because rather than simply listing the models attributes, it gets attributes from a model. More of a function than a type really.

RobinBuschmann:
@Danlock I would be more happy with `ModelAttributes` or so. But lets do this later. I will merge it. Thank you very much :)

Danlock:
Maybe ParsedModelAttributes or FilteredModelAttributes? Thanks for merging.

RobinBuschmann:
`FilteredModelAttributes` sounds good :)

PR #253, opened 2017-12-28
Export ICountOptions and use Generics.

santiagodoldan:


bilby91:
@RobinBuschmann Are you fine with this change ?

lib/models/Model.d.ts
```diff
@@ -234,7 +234,7 @@ export declare abstract class Model<T> extends Hooks {
    *
    * If you provide an `include` option, the number of matching associations will be counted instead.
    */
-  static count(options?: ICountOptions): Promise<number>;
+  static count<T extends Model<T>>(options?: ICountOptions<T>): Promise<number>;
```
RobinBuschmann: Can you add `this: (new () => T)` here as well? See `findAll` as an example: https://github.com/suttna/sequelize-typescript/blob/42240661d503d614781d1353f131a37db33300be/lib/models/Model.d.ts#L201

RobinBuschmann:
@santiagodoldan Thanks for contributing @santiagodoldan @bilby91 There is one thing left - see the my comment :)

santiagodoldan:
@RobinBuschmann done, thanks to you ☺️

PR #255, opened 2017-12-29
Allow passing the scopes as consecutive arguments

omidkrad:
This is documented [here](http://docs.sequelizejs.com/manual/tutorial/scopes.html#merging).

> Several scopes can be applied simultaneously by passing an array of scopes to .scope, or by passing the scopes as consecutive arguments.

```JavaScript
// These two are equivalent
Project.scope('deleted', 'activeUsers').findAll();
Project.scope(['deleted', 'activeUsers']).findAll();
```


RobinBuschmann:
@omidkrad Thanks for contributing :)

Issue #252, opened 2017-12-26
missing dataValues property in Modal.d.ts


santiagodoldan:
Just wanted to know if there's a reason or if it is ok to add the missing `dataValues` property, I'd love to add it as we're using it to get the plain object of a model instance.

https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/models/Model.d.ts

RobinBuschmann:
@santiagodoldan Sounds good. I have nothing against it :)

PR #257, opened 2018-01-04
Add dataValues to Model.d.ts

santiagodoldan:


santiagodoldan:
https://github.com/RobinBuschmann/sequelize-typescript/issues/252

santiagodoldan:
@RobinBuschmann is this ok? at least it works as expected 😂

RobinBuschmann:
Hey @santiagodoldan, thanks for implementing. Since the last merge we do have a type called `GetAttributes<T>`, which omits the methods from the Model Interface and does only provide what comes from T and `{id?: number|any; createdAt?: Date|any; updatedAt?: Date|any; deletedAt?: Date|any; version?: number|any;}` I think this would be a more suitable interface.What do you think? :)

santiagodoldan:
Got these error messages `Property 'id' does not exist on type 'GetAttributes<User>'. Property 'id' does not exist on type 'Partial<Omit<User, "id" | "$add" | "$set" | "$get" | "$count" | "$create" | "$has" | "$remove" | ...'` or `Property 'customAttr' does not exist on type 'GetAttributes<User>'. Property 'customAttr' does not exist on type '{ id?: any; createdAt?: any; updatedAt?: any; deletedAt?: any; version?: any; }'.`

santiagodoldan:
We're using `@Column` to define them, could be that an issue?

```
  @PrimaryKey
  @Column({ autoIncrement: true })
  public id: number
```

RobinBuschmann:
@santiagodoldan Hmm... good questions. I couldn't find out yet. I will come back to you, when I know what the problem is.

RobinBuschmann:
@santiagodoldan Ok, should be fixed now. Additionally I've renamed the type. Can you give it another try? :)

santiagodoldan:
@RobinBuschmann yep, works ok now, thanks

PR #264, opened 2018-01-09
Update README.md

mossaabzaroui:
@RobinBuschmann 

Fix file extension for model import.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,10 +218,10 @@ import {Sequelize} from 'sequelize-typescript';
 
 const sequelize =  new Sequelize({
         ...
-        modelPaths: [__dirname + '/**/*.model.ts']
+        modelPaths: [__dirname + '/**/*.model.js']
 });
 // or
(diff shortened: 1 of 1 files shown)


RobinBuschmann:
Hey @mossaabzaroui thanks for contributing, but actually `*.ts` is not wrong, when using `ts-node`, `ts` files can be loaded directly. Probably mention both (usage with `*.js` AND `*.ts`) and making a hint to `ts-node` would make sense. What do you think?

mossaabzaroui:
@RobinBuschmann 
- Are you suggesting to change ts-config to point Typescript which extention file should be used instead of changing the current modelsPath ?

- When i use .ts extension i have this error : Model not initialized: "<Model Name>" needs to be added to a Sequelize instance before "create" can be called. Did you have any idea about this issue ?

RobinBuschmann:
Regarding Your last point: can you provide an example?

RobinBuschmann:
Close this issue due to no activity

PR #269, opened 2018-01-16
fix Model.changed method arguments description

vologab:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3382


RobinBuschmann:
Hey @vologab, thanks for contributing. Can you provide a link where this is documented? Thank you :)

vologab:
Hi @RobinBuschmann. I've put link to sequelize repo inside commit message:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L3382.
Also issue when we have to use second param: https://github.com/sequelize/sequelize/issues/2862

RobinBuschmann:
Sry, I have overseen that.

PR #280, opened 2018-01-22
validateOnly property type was changed to boolean as expected

vlavrynovych:


RobinBuschmann:
Hey @vlavrynovych, thanks for contributing, but it is as it should be :) When you set `validateOnly` to `true` only `modelPaths` and the extended members (from `Options`) should be available (`ISequelizeValidationOnlyConfig `). When setting it to `false` all other interfaces are available. It helps the typescript compiler to infer the correct interface.

vlavrynovych:
Hello @RobinBuschmann, it was really unclear to me. Sorry for that and thank you for explanation.

lookapanda:
@RobinBuschmann This `validateOnly` is causing problems for me. The compiler does not want to accept this:
```js
const sequelize = new Sequelize({
            database: credentials.database,
            dialect: 'mysql',
            username: credentials.username,
            password: credentials.password,
            replication: {
                write: {
                    host: credentials.write.host,
                },
                read: [
                    {
                        host: credentials.read.host,
                    },
                ],
            },
            dialectOptions: {
                bigNumberStrings: true,
            },
            pool: {
                max: 10,
            },
            validateOnly: false,
        });
```

For some miraculous reason, it does accept it when I remove the `read` property from the config.


bildschirmfoto 2018-05-30 um 15 53 58

lookapanda:
Could it be because `sequelize-typescript` is using `@types/sequelize` in version `4.27.14`? Because it could be that in that version `read` was not an array (as I use it), but in a more up2date version, it is! 

Could we update that typings please?

PR #293, opened 2018-02-06
Use Partial<T> for set and update keys

omidkrad:
This allows typescript to check only the model properties are passed to `set` and `update` methods.

### DIFF
--- a/lib/models/Model.d.ts
+++ b/lib/models/Model.d.ts
@@ -560,10 +560,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
    * @param options.reset Clear all previously set data values
    */
   set(key: string, value: any, options?: InstanceSetOptions): this;
-  set(keys: Object, options?: InstanceSetOptions): this;
+  set(keys: Partial<T>, options?: InstanceSetOptions): this;
 
   setAttributes(key: string, value: any, options?: InstanceSetOptions): this;
-  setAttributes(keys: Object, options?: InstanceSetOptions): this;
+  setAttributes(keys: Partial<T>, options?: InstanceSetOptions): this;
 
   /**
    * If changed is called with a string it will return a boolean indicating whether the value of that key in
@@ -612,10 +612,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
    * This is the same as calling `set` and then calling `save`.
    */
   update(key: string, value: any, options?: InstanceUpdateOptions): Promise<this>;
-  update(keys: Object, options?: InstanceUpdateOptions): Promise<this>;
+  update(keys: Partial<T>, options?: InstanceUpdateOptions): Promise<this>;
 
   updateAttributes(key: string, value: any, options?: InstanceUpdateOptions): Promise<this>;
-  updateAttributes(keys: Object, options?: InstanceUpdateOptions): Promise<this>;
+  updateAttributes(keys: Partial<T>, options?: InstanceUpdateOptions): Promise<this>;
 
   /**
    * Destroy the row corresponding to this instance. Depending on your setting for paranoid, the row will
--- a/lib/models/Model.d.ts
+++ b/lib/models/Model.d.ts
@@ -560,10 +560,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
    * @param options.reset Clear all previously set data values
    */
   set(key: string, value: any, options?: InstanceSetOptions): this;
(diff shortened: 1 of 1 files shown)


lib/models/Model.d.ts
```diff
@@ -560,10 +560,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
    * @param options.reset Clear all previously set data values
    */
   set(key: string, value: any, options?: InstanceSetOptions): this;
-  set(keys: Object, options?: InstanceSetOptions): this;
+  set(keys: Partial<T>, options?: InstanceSetOptions): this;
```
RobinBuschmann: `T` need to be `this` :)

lib/models/Model.d.ts
```diff
@@ -560,10 +560,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
...
 
   setAttributes(key: string, value: any, options?: InstanceSetOptions): this;
-  setAttributes(keys: Object, options?: InstanceSetOptions): this;
+  setAttributes(keys: Partial<T>, options?: InstanceSetOptions): this;
```
RobinBuschmann: same here..

lib/models/Model.d.ts
```diff
@@ -612,10 +612,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
    * This is the same as calling `set` and then calling `save`.
    */
   update(key: string, value: any, options?: InstanceUpdateOptions): Promise<this>;
-  update(keys: Object, options?: InstanceUpdateOptions): Promise<this>;
+  update(keys: Partial<T>, options?: InstanceUpdateOptions): Promise<this>;
```
RobinBuschmann: ...here...

lib/models/Model.d.ts
```diff
@@ -612,10 +612,10 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
...
 
   updateAttributes(key: string, value: any, options?: InstanceUpdateOptions): Promise<this>;
-  updateAttributes(keys: Object, options?: InstanceUpdateOptions): Promise<this>;
+  updateAttributes(keys: Partial<T>, options?: InstanceUpdateOptions): Promise<this>;
```
RobinBuschmann: ...and here ;)

RobinBuschmann:
Thanks for contributing :) After adjusting the typings it should compile

omidkrad:
Failed again. I don't think the problem was `Parial<T>` actually it makes more sense for the `keys` to be of type `Partial<T>` rather than `Partial<this>` because properties of the base model are probably not valid keys.

RobinBuschmann:
`this` is not pointing to the base model here, but 

> `this` type represents a type that is the subtype of the containing class or interface
> https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types

The following example makes it very clear why we should use `this` over `T`:

### Using `this`
```ts
class Model<T> {
  dataValues: FilteredModelAttributes<this>; // <-- this !
}
class Parent extends class Model<Parent> {
  @Column
  a: string;
}
class Child extends Parent {
  @Column
  b: string;
}

const child: Child = createChildInstance(); // doesn't matter how we've created an instance here

child.dataValues.a // works as expected
child.dataValues.b // works, because the generic type in "FilteredModelAttributes" points to "this", "b" is available here
```

### Using `T`
```ts
class Model<T> {
  dataValues: FilteredModelAttributes<T>; // <-- T !
}
class Parent extends class Model<Parent> {
  @Column
  a: string;
}
class Child extends Parent {
  @Column
  b: string;
}

const child: Child = createChildInstance(); // doesn't matter how we've created an instance here

child.dataValues.a // works as expected
child.dataValues.b // <-- Error "b" does not exist on "T". Because "T" is "Parent" not "Child"
```

RobinBuschmann:
The tests are failing correctly, hence the types a stricter now:
```
test/specs/instance.spec.ts(1550,48): error TS2345: Argument of type '{ validateTest: string; }' is not assignable to parameter of type 'Partial<User>'.
  Object literal may only specify known properties, and 'validateTest' does not exist in type 'Partial<User>'.
1550         .then((user) => user.updateAttributes({validateTest: 'hello'}))
```

As the error message say, `validateTest` is no property of `User`. Everything works as intended :) So great work!

What's left to make this PR build is adjusting the tests.

omidkrad:
@RobinBuschmann I'm not sure if I will get time to go through these tests, please feel free to make the changes if you think this PR is correct.

PR #313, opened 2018-02-23
Added sequelize.where to IBaseIncludeOptions 

ahmedshaaban:
as there was compiling error when writing a find function line this `return await PublishedRoute.findAll({
            include: [{
                model: Route,
                where: sequelize.where(sequelize.col("PublishedRoute.versionId"), "=",
                    sequelize.col("routes.id")),
                include: [Stop]
            }]
        });`


RobinBuschmann:
@ahmedshaaban thanks for contributing!

PR #314, opened 2018-02-24
Add static increment method to Model

### LINKED ISSUE #276: Static increment method is missing from Model
Currently doing:
```ts
const await user = User.findById(1);
user.increment({ amount });
```

Would be nice to not need to do the first query. Looked through the documentation/code and found that there's actually a static increment as well

http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-increment
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2929

intellix:
Fixes #276 

Originally I thought there was also a static decrement method but seems it's missing. Might be nice to add that upstream to Sequelize afterwards.

One thing I was confused about was why the `InstanceIncrementDecrementOptions` doesn't contain silent and it's extended afterwards. Guess I'll just leave it as it was for now. The description for the static increment was copied across from Sequelize too

Return type depends on if you pass in returning for Postgres: https://github.com/sequelize/sequelize/blob/master/lib/model.js#L2999

I did originally have `Promise<[Array<T>, number] | Promise<[number]>` but TS 2.7 requires fixed length tuples

### DIFF
--- a/lib/models/Model.d.ts
+++ b/lib/models/Model.d.ts
@@ -404,6 +404,27 @@ export declare abstract class Model<T extends Model<T>> extends Hooks {
   static restore(options?: RestoreOptions): Promise<void>;
 
   /**
+   * Increment the value of one or more columns. This is done in the database, which means it does not use the values currently stored on the Instance. The increment is done using a
+   * ```sql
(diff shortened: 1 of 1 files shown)


RobinBuschmann:
👍 thank you

PR #337, opened 2018-03-24
Update Model.scope definition

doovers:
Allow simple scopes and scope functions together. For example:

```typescript
Product
  .scope('defaultScope', { method: ['company', companyId] })
  .find();
```


doovers:
Oops! Just realized that someone beat me to it with a better solution!

PR #345, opened 2018-04-05
Repository mode

alejandrogkl:


lib/models/BaseSequelize.ts
```diff
@@ -19,21 +20,25 @@ import {BaseAssociation} from './association/BaseAssociation';
...
   }
 
   static extend(target: any): void {
-
+    const _addModels = target.prototype.addModels;
     extend(target, this);
+    if (_addModels) {
+      target.prototype.addModels = _addModels;
+    }
   }
```
RobinBuschmann: Can you explain the purpose of this? Is there a case where it can be happen that `addModels` doesn't exist anymore after `extend` is called?

lib/interfaces/ISequelizeConfig.ts
```diff
@@ -1,7 +1,6 @@
-import {Options} from 'sequelize';
+import { Options } from 'sequelize';
```
RobinBuschmann: In general: can you revert these style changes everywhere? We're aren't using spaces on the start and the end in `{}` Thank you :)

lib/models/v4/Sequelize.ts
```diff
@@ -27,33 +36,48 @@ export class Sequelize extends OriginSequelize implements BaseSequelize {
...
+  addModels(modelPaths: string[]): void;
+  addModels(arg: Array<typeof Model | string>): void {
+    let models = getModels(arg);
+    if (this.repositoryMode) {
+      models = models.map(model => {
+        this._repos[model.name] = ModelFactory(model);
+        return this._repos[model.name];
+      });
+    }
+    //
```
RobinBuschmann: Remove unnecessary `//`

lib/models/BaseSequelize.ts
alejandrogkl: i implement the addModels function in v4 to take into account the repositoryMode, but the extend override the v4 function with the one on BaseSequelize, so if the v4/v3 have that function implemented i keep that one and not the one on BaseSequelize.

the reason to implement the addModels in v4 and not change the one on BaseSequelize, was because the project in general use es5 and then this code

```
export function ModelFactory(model: any): any {
  return class extends model {};
}
```

compile into something that doesn't work, ( no classes in es5 ), so i decided to implement that in Sequelize v4 since that file is compiled to es6

When you drop support of v3, that patch could be removed.

If you have a better way to create the derived class of the model that work in es5 , let me know, so i can remove the addModels from v4 and modified the one on BaseSequelize.

lib/interfaces/ISequelizeConfig.ts
alejandrogkl: ok not problem, i will disable my formatter and create another pull request. is that ok?

lib/models/v4/Sequelize.ts
alejandrogkl: ok

lib/interfaces/ISequelizeConfig.ts
RobinBuschmann: The changes will take place automatically on this PR already - So no need for creating another one :)

lib/models/BaseSequelize.ts
alejandrogkl: definitely a better idea

alejandrogkl:
HI,
Any plans to merge the changes.
Do you need something else?

Please let me know.

RobinBuschmann:
Hey @alejandrogkl , yes, we definitely need tests for that. If you need help with this let me know :)

alejandrogkl:
I do, can you give some guidance of which can of test should i write.

RobinBuschmann:
@alejandrogkl 
Regarding **unit testing**:
Init a repository model and the same model without repository mode,
then compare the outcome: After initialization the model classes will have the following properties among others:
  - `rawAttributes`
  - `associations`
  - `tableName`
  - `primaryKeys`

I think comparing these would be reasonable.

For **integration tests**, it would make sense to test (one very simple example per keyword will be enough):
- validations
- hooks
- associations
- whether one model can be used with multiple sequelize instances

A completely new test file would be fine (`repository-mode.spec`)

What do you think?

RobinBuschmann:
@alejandrogkl But I can also fork your branch to support you directly, if you like to :)

alejandrogkl:
i make a commit with the tableName test, just to check if i understood you properly.
You can definitely fork my branch. i will appreciate your help

alejandrogkl:
HI @RobinBuschmann 

Please let me know if the tableName test is correct to write a few more.

RobinBuschmann:
@alejandrogkl sry for the late reply. Yes, the test looks good :) but i’m Not sure about the other thing you’ve changed. Unfortunately I hadn’t time yet to take a closer look at it.

lib/models/association/BaseAssociation.ts
```diff
@@ -8,8 +8,8 @@ import {ModelClassGetter} from '../../types/ModelClassGetter';
 import {ModelNotInitializedError} from '../errors/ModelNotInitializedError';
```
RobinBuschmann: I suggest reverting all changes here and move all repository mode related stuff to https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/models/BaseSequelize.ts#L104 Then only the sequelize instance need to care about repository stuff.

What do you think?

620eli:
Any updates on this one?

620eli:
I used the PR in my project. Everything working fine except the scopes which include models because you can't use a sequelize instance in a model class.

alejandrogkl:
Hi @620eli  could you post an example of your use case. Maybe i can find a temporal solution. I am also using the PR in a project since May.

@RobinBuschmann  i couldn't find a way to isolate the repositoryMode from the BaseAssociation, the main issue was that the getAssociatedClass in BaseAssociation is getting call from many places and they expect an initialized class, so i had to interfiere and return the initialized one from the sequelize connection. It was the quickest way, and since the BaseAssociation class was already receiving the sequelize connection in the init function, i just saved the connection in a variable and used the variable inside other functions.

Do you any idea of who to archive the independence of the BaseAssociation, without a major code refactor?

Do you have an estimation of when 1.0.0 will be released?

Thanks for this awesome library.

620eli:
@alejandrogkl Well, the simplest solution to that problem is to just not use scopes. Including the models inside the query options seems to be working fine for now.

alejandrogkl:
@620eli could you post an example of your use case?

Your solution also work :), but maybe is something easy to fix, and then you can advantage of what the scope provide and avoid code repetition.

RobinBuschmann:
Hey @alejandrogkl, I've released a new alpha version of sequelize-typescript (`1.0.0-alpha.4`) - It includes a preparation for the repository mode... I also have a better understanding now on how to achieve the repository mode. I'll come back with my ideas very soon. Probably this will help you implementing it for 0.6.X

RobinBuschmann:
Repository mode is implemented on branch `1.0.0` and released with `sequelize-typescript@1.0.0-alpha.7`

PR #351, opened 2018-04-10
export validator annotation

kutase:
Fix validator export.


RobinBuschmann:
Thanks for implementing :)

Issue #330, opened 2018-03-19
Typescript 2.7.2 constructor function


sanewell92:
Got an error, originating from the sequelize type definitions.

```
node_modules/sequelize-typescript/lib/models/Sequelize.d.ts(7,40):
  error TS2507:
  Type '{
    default: SequelizeStatic;
    cls: any;
    useCLS(namespace: Namespace): Sequelize;
    Utils: Utils;
    Prom...'
  is not a constructor function type.
```

Coming from `Sequelize.d.ts`.

Anyone else getting this error? I'm using typescript 2.7.2 on an Ubuntu machine.

RobinBuschmann:
Hey @sanewell92 thanks for reporting. Unfortunately I cannot reproduce the issue. Can you provide an example repo which demonstrates the problem? Thank you

sanewell92:
Try cloning [this repo](https://github.com/sanewell92/test-sequelize-ts-2.7) and then `npm install && npm run build`

RobinBuschmann:
@sanewell92 thanks for creating this repo. The compiler option `esModuleInterop` is causing this error. This is because `esModuleInterop: true` requires exports like `export =` to be imported with `import target from 'source'`. Since the sequelize typings use this format (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sequelize/index.d.ts#L6620) it is throwing this error. 
So for now, either omit this compiler option (`esModuleInterop`) or set `skipLibCheck: true` (see https://www.typescriptlang.org/docs/handbook/compiler-options.html).

But for the long term I will need to investigate this a little bit more and check how I will proceed with this issue.

rochapablo:
`skipLibCheck: true` worked, but when I try so do `.save()` I'm getting `this.lib.Database is not a constructor`

rochapablo:
Oops, I guess that it's because I was using `validateOnly: true,`. Still I'm getting access denied which it's odd, checking...

junior-ales:
I've spent a good amount of time dealing with this compilation error of `SequelizeOrigin` in `Sequelize` class. Maybe we could add something on the installation instructions about the `esModuleInterop` option, I'm up for opening a PR if you think it is appropriate @RobinBuschmann.

RobinBuschmann:
@junior-ales Sounds good :) For `sequelize-typescript@1.0.0` this won't be an issue anymore. But for now it is a good idea.

PR #360, opened 2018-04-24
Use esModuleInterop and fix corresponding issues

johannes-scharlach:
Fixes https://github.com/RobinBuschmann/sequelize-typescript/issues/330

I've also tried [bumping `typescript` and setting `esModuleInterop` to `true`](https://github.com/johannes-scharlach/sequelize-typescript/commit/9c5daa47f079c48ebe9d5eeaee87a0c65344e1e7). The proposed changes also worked with the newer version, but I'm not sure if that would introduce a breaking change to `sequelize-typescript`. The recommendation seems to be to set that flag to true when you can.

I've taken the solution from this issue https://github.com/Microsoft/TypeScript/issues/21535#issuecomment-362399044

### DIFF
--- a/lib/models/Sequelize.d.ts
+++ b/lib/models/Sequelize.d.ts
@@ -1,5 +1,5 @@
 import 'reflect-metadata';
-import * as SequelizeOrigin from 'sequelize';
+import SequelizeOrigin = require('sequelize');
 import {Model} from "./Model";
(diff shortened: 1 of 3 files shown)


RobinBuschmann:
I just published a new release including this change: `sequelize-typescript@0.6.5-beta.0`

PR #362, opened 2018-04-25
Set freezeTableName in global options and use those with all configs

johannes-scharlach:
Technically a breaking change is that global option defaults were not set if you would call `new Sequelize` with a `ISequelizeUriConfig` or just a uri. This behaviour has changed now, so some people may have had timestamps so far and won't get them anymore once this MR is published.

Technically I think this is a bug, as sequelize-typescript documents the behaviour without no timestamps by default. (Another difference with sequelize.)

I don't have any urgency to get this merged, so you may want to make this part of v1.0.0? Let me know if I should change the MR to target that branch.

### DIFF
--- a/lib/annotations/Table.ts
+++ b/lib/annotations/Table.ts
@@ -19,9 +19,6 @@ export function Table(arg: any): void|Function {
 }
 
 function annotate(target: any, options: IDefineOptions = {}): void {
-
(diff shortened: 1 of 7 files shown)


RobinBuschmann:
Perfect, thanks

PR #363, opened 2018-04-25
WIP: allow setting unique constraint keys in @Unique decorator

johannes-scharlach:
I was looking for #176 a while ago and have this written up. Is that what you had in mind? I'd be happy to add some tests in order to get this merged.

### DIFF
--- a/lib/annotations/Unique.ts
+++ b/lib/annotations/Unique.ts
@@ -1,12 +1,27 @@
 import 'reflect-metadata';
 import {addAttributeOptions} from "../services/models";
 
-/**
- * Sets unique option true for annotated property.
- */
-export function Unique(target: any, propertyName: string): void {
+export type uniqueAnnotation = (target: any, propertyName: string) => void;
 
-  addAttributeOptions(target, propertyName, {
+/**
+ * Sets unique option for annotated property.
+ */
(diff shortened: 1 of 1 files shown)


lib/annotations/Unique.ts
```diff
@@ -1,12 +1,27 @@
 import 'reflect-metadata';
 import {addAttributeOptions} from "../services/models";
 
+export type uniqueAnnotation = (target: any, propertyName: string) => void;
```
RobinBuschmann: Can you use capital letters here?

lib/annotations/Unique.ts
```diff
@@ -1,12 +1,27 @@
...
+
 /**
- * Sets unique option true for annotated property.
+ * Sets unique option for annotated property.
  */
-export function Unique(target: any, propertyName: string): void {
+export function Unique(keyCombinator: string): uniqueAnnotation;
```
RobinBuschmann: I think we should provide all possible options here (see: https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/interfaces/IPartialDefineAttributeColumnOptions.ts#L18) What do you think?

RobinBuschmann:
Besides the mentioned issue, tests would be good to make ci happy :)

PR #374, opened 2018-05-28
added overloading for upsert and insertOrUpdate

donaldpipowitch:
Hi there,

I'm not super familiar with `sequelize-typescript`, but I noticed that `upsert`/`insertOrUpdate` actually return a tuple inside a Promise, when `return: true` is used. This can probably be typed correctly with overloading.

In my case A is a model. Not sure if this covers all cases, though.


RobinBuschmann:
Great! Thanks for fixing.

PR #409, opened 2018-07-12
Added duplicating property to IBaseIncludeOptions

francisvalois:
This property allow a choice to avoid a cartesian product with assocations when using include. The actual Sequelize code is located at the line 487 of model.js. Also, this property in mentioned in multiple PR and Issues on Sequelize repo.


RobinBuschmann:
Sry for the delayed reply. Looks good! Thanks for fixing

PR #419, opened 2018-07-26
Added custom model matching

### LINKED ISSUE #405: Custom filename -> model matching
Would it be possible to allow a custom matcher instead of matching the filename directly to the model? I use `person.model.ts` as filename and `export class Person extends Model<Person>`. I don't want to use default exports because of reasons mentioned here: https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html.

I would suggest adding an optional modelMatch property to the Sequelize constructor. This property would be given a function which would be supplied with the filename (without extension) of a module that matches the modelPath and its exported members. The function should for each exported member decide (true or false) whether it is a model. 

**example**

```TypeScript
// person.model.ts

export interface PersonResponse {
}

export class Person extends Model<Person> {
}
```

```TypeScript
const sequelize = new Sequelize({
    modelPaths: [path.join(__dirname, 'api/**/*.model.ts')],
    modelMatch: (filename, member) => filename.startsWith(member.toLowerCase())
});
```

In this case, the modelMatch function would be called twice, one with `('person.model.ts', 'PersonResponse')` and once with `('person.model.ts', 'Person')`. As soon as this function return true, the matched exported member should be added as model.

Consumers of this library can then use any desired approach to matching, be it regexes, a predefined dictionary, ...

The default for this function would be `(filename, member) => filename === member`.

If you're interested in this proposal, I can make a PR.

MatthiasKunnen:
Added model matching as described in #405. Also fixed two issues I came across:

- Double extension stripping in getModels
- Cross test pollution of the sequelize instance in `sequelize.spec`

Resolves #405

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,14 +238,61 @@ sequelize.addModels([__dirname + '/**/*.model.ts']);
 ```
 
 #### Model-path resolving
-When using a path to resolve the required models, either the class has to be exported as default or if not exported
-as default, the file should have the same name as the corresponding class:
+A model is matched to a file by its filename. E.g.
 ```typescript
-export default class User extends Model<User> {}
-
-// User.ts
+// File User.ts matches the following exported model.
 export class User extends Model<User> {}
 ```
+This is done by comparison of the filename against all exported members. The
+matching can be customized by specifying the `modelMatch` function in the
+configuration object.
+
+For example, if your models are named `user.model.ts`, and your class is called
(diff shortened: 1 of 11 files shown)


RobinBuschmann:
Looks great 👍

RobinBuschmann:
Its released now with `sequelize-typescript@0.6.6`

PR #425, opened 2018-07-29
README: Minor fix

pyk:


RobinBuschmann:
Thanks for contributing and sry for the late reply

PR #430, opened 2018-08-01
SQL syntax error. findAndCountAll with scopes

ALjean:
allowed  sequelize findAndCountAll options set empty include array.
It's default sequelize behaviour.


RobinBuschmann:
@ALjean thanks for implementing. Can you also add a test which covers the issue you actually had?

ALjean:
@RobinBuschmann Yes. it's still WIP

PR #444, opened 2018-08-28
add limit option to include options

ahmedshaaban:
if i want the first element only in the array , for example something like the following code
```
Student.findOne({
            where: { uuid },
            include: [{
                model: Stop,
                include: [
                    {
                        model: RoundStop,
                        where: {
                            arrivesAt: {
                                [Op.gte]: time
                            }
                        },
                        // @ts-ignore
                        limit: 1,
                        required: false,
                        order: [["arrivesAt", "ASC"]],
                        include: [Round]
                    }
                    , Route]
            }]
        });```


RobinBuschmann:
@ahmedshaaban Thanks for fixing. 👍

Issue #446, opened 2018-08-30
Still creating invalid Postgres/MySQL queries


TimMensch:
Bug #135 was closed for "no followup", but is being ignored even though another person posted that he was having the same issue.

Well, now I'M having this issue, and I tracked it down to a bug in `sequelize-typescript`. Rather that post it on the closed bug, where based on the history it may be ignored, I'm filing a new bug.

It's a combination of three things:

* sequelize-typescript sets `distinct:true` as a default.
* sequelize-typescript deletes an empty `include:[]` array from settings.
* sequelize-typescript then calls `count()` on the model, which causes Sequelize to generate bad SQL.

Leaving an empty `include` array fixes the problem. This can be fixed by deleting a single line from `/lib/services/models.js`:

```javascript
/**
 * Pre conform includes, so that "as" value can be inferred from source
 */
function inferAlias(options, source) {
    options = __assign({}, options);
    if (!options.include) {
        return options;
    }
    // if include is not an array, wrap in an array
    if (!Array.isArray(options.include)) {
        options.include = [options.include];
    }
    else if (!options.include.length) {
//        delete options.include; // *** Removing this line fixes the issue, and it seems redundant anyway
        return options;
    }
    // convert all included elements to { model: Model } form
    options.include = options.include.map(function (include) {
        include = inferAliasForInclude(include, source);
        return include;
    });
    return options;
}
```

To repro the bug, you should be able to create any table with sequelize-typescript, and then share the service with pagination enabled (using `feathers-sequelize`):

```javascript
    const userService = service({ Model: User, paginate: { max: 50, default: 50 } });
    app.use("users", userService);    
```

`feathers-sequelize` calls findAndCountAll() on the model. `feathers-sequelize` defaults the `distinct` to true, and while there's a workaround that involves adding a hook to every service, it would be nice if this combination didn't crash at the `sequelize-typescript` layer.

I'm currently not set up to perform pull requests on sequelize-typescript; please let me know if you can apply this fix ASAP so that we don't need to continue work around this bug.

TimMensch:
(Updated the bug report to clarify that I was using `sequelize-typescript` with `feathers-sequelize`.)

RobinBuschmann:
@TimMensch Thanks for reporting. Can you create a PR for that?

TimMensch:
OK, PR is done.

TimMensch:
[...]

PR #450, opened 2018-09-04
Fix the bug described in #446

TimMensch:
* Delete a line in `inferAlias` that removes an empty `includes` object. Deleting this object actually can make Sequelize produce illegal SQL code.

Bug #446 describes this issue. The fix is trivial and seems harmless. All tests pass.

### DIFF
--- a/lib/services/models.ts
+++ b/lib/services/models.ts
@@ -260,7 +260,6 @@ export function inferAlias(options: any, source: any): any {
   if (!Array.isArray(options.include)) {
     options.include = [options.include];
   } else if (!options.include.length) {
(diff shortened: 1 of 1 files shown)


TimMensch:
I'm going to assume that, by deleting a line and therefore changing the "coverage" by 0.01%, you're going to ignore that "failed" test, yes? Because I'd love to see this fix get merged, but I don't have time to add tests to unrelated code in order to "fix" the coverage.

RobinBuschmann:
@TimMensch Yeah, i will ignore the coverage report. Thanks for fixing!

PR #476, opened 2018-10-24
added findByPk method to the Model

tzellman:
findById and findByPrimary are now deprecated
changes in sequelize: https://github.com/sequelize/sequelize/commit/1bc8b0e82d4635ff22cbd4c048c042fed30a036f

I wasn't sure if I should also push a PR to https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/sequelize, as the `findByPk` method was just added a few days ago (in above referenced commit), so if users aren't using the latest 4.x sequelize version.


RobinBuschmann:
@tzellman Thanks for adding. 

The sequelize typings doesn't affect sequelize-typescript in that case

PR #481, opened 2018-11-15
Feature/#308 dollar operators

RobinBuschmann:


rajatsehgal01:
Now is there any operator feature available in the any version of sequelize-typescript?

PR #484, opened 2018-11-21
Added the paranoid flag to the ICountOptions interface

rokerkony:
Hi there, I added the missing paranoid flag to count options.

Now this is supported:
```ts
return Entity.count({
  paranoid: false,
});
```


RobinBuschmann:
Thanks :)

PR #488, opened 2018-12-03
Added missing hooks property for ICreateOptions

jackphilippi:
`hooks` property for ICreateOptions was missing as per the spec at http://docs.sequelizejs.com/class/lib/model.js~Model.html#static-method-create. Rest of the properties seem to be fine. 

Thanks!


RobinBuschmann:
@jackphilippi Thanks for contributing!

PR #491, opened 2018-12-10
Add nest option for find

CSLTech:
While poorly documented find options derive from query options which include the nest option. Added it in as it's a pretty usefull option to have.


RobinBuschmann:
@CSLTech sry for the late reply.Should that be a part of the `QueryOptions` instead like in the official typings of sequelize: https://github.com/sequelize/sequelize/blob/26547bc92ebfe235507092d63020310256f43850/types/lib/query-interface.d.ts#L32

RobinBuschmann:

Issue #244, opened 2017-12-18
Model.$set definitions conflicts with implementation


damianb:
As above.  Typescript is currently throwing a fit, because...

[This](https://github.com/RobinBuschmann/sequelize-typescript/blob/0f7e1d450a975be1fdf2cf24165b65799bfe0493/lib/models/Model.d.ts#L428) =/= [this](https://github.com/RobinBuschmann/sequelize-typescript/blob/ddc0608c902c42b496c6c213d7178007ab610669/lib/models/BaseModel.ts#L133)

Right now, Typescript is insisting that the third parameter for $set must conform to the interface IAssociationActionOptions, which both conflicts with the implementation *and* prevents passing `save: false`, which is useful when creating multiple associations while disallowing null values.

RobinBuschmann:
@damianb So `save: boolean` is missing in `IAssociationActionOptions`, right? Can you provide a link to the docs describing these options? I couldn't find them. Thank you :)

damianb:
Not sure if it's that it is missing from the interface, or if the interface itself for $set's options is distinct in and of itself.

This is what I was looking at:

http://docs.sequelizejs.com/class/lib/associations/belongs-to.js~BelongsTo.html#instance-method-set

RobinBuschmann:
@damianb Sry for the late response. I found the set methods in the sequelize docs: http://docs.sequelizejs.com/class/lib/associations/has-many.js~HasMany.html#instance-method-set

Unfortunately it doesn't seem to have a `save` option. Can you confirm that it is working anyway with `save: false`? In this case, we could add it to the typings anyway. Thank you

damianb:
Iiiinteresting...looking at the sequelize source, it seems you're right for the rest.

Though, looking at the BelongsTo definition, it *should* still work in my case; I'm trying to set their relationship via the xref table (because I have an additional column that needs set...), and I'm using BelongsTo like below:

```
export class UsergroupPermission extends Model<UsergroupPermission> {
  @ForeignKey(() => Permission)
  @AllowNull(false)
  @Column({
    unique: 'UserGroupPermission'
  })
  public PermissionId: number

  @ForeignKey(() => Usergroup)
  @AllowNull(false)
  @Column({
    unique: 'UserGroupPermission'
  })
  public UsergroupId: number

  @AllowNull(false)
  @Column({
    type: DataType.ENUM(...PermissionValueArray),
    comment: 'The value of the specified permission for the given UserGroup.'
[...]

PR #493, opened 2018-12-13
fix typing definition of Model.$set

starcharles:
current typing definition for Model.$set method  seems to be wrong.

docs: http://docs.sequelizejs.com/class/lib/associations/belongs-to.js~BelongsTo.html#instance-method-set

This problem is also mentioned here.
https://github.com/RobinBuschmann/sequelize-typescript/issues/244#issuecomment-413635439


RobinBuschmann:
Fixed with #508

PR #502, opened 2018-12-26
[Readme.md] update the v4 command

piyushkantm:
Update the installation command to correctly install v4 for sequelize instead of installing v5beta


RobinBuschmann:
Hey @piyushkantm, thanks for contributing, but this isn't necessary for now... since the latest sequelize version is 4.X, `npm i sequelize` installs a 4er version

PR #506, opened 2019-01-10
Update ISequelizeConfig

jlberrocal:
username and password should be non required fields, i.e. SQLite doesn't require those fields

### DIFF
--- a/lib/interfaces/ISequelizeConfig.ts
+++ b/lib/interfaces/ISequelizeConfig.ts
@@ -11,12 +11,12 @@ export interface ISequelizeConfig extends Options {
   /**
    * Username of database
    */
-  username: string;
+  username?: string;
 
   /**
    * Password for database user
    */
-  password: string;
+  password?: string;
 
   /**
    * Path to models, which should be loaded


RobinBuschmann:
Hey @jlberrocal, thanks for contributing, but this should already be done with https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/interfaces/ISequelizeStorageConfig.ts. Or do I overseen something?

jlberrocal:
@RobinBuschmann not exactly, in that way i can't set the dabase name nor loggin properties, and also right now `Sequelize` constructor doesn't allow `ISequelizeStorageConfig` it is expecting for `SequelizeConfig | ISequelizeValidationOnlyConfig` or a `string` url

RobinBuschmann:
Hey @jlberrocal `SequelizeConfig` is a union type, which consists of `ISequelizeStorageConfig`. Since `ISequelizeStorageConfig` extends `Options` (`import {Options} from 'sequelize';`) you should be definitely able to set all other options.

jlberrocal:
That's weird, on my definition file (node_modules) I can only see ISequelizeConfig or string, I tried with storege but is being marked as error

RobinBuschmann:
@jlberrocal Which sequelize-typescript version are you using? As you can see [here](https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/types/SequelizeConfig.ts#L6)  `SequelizeConfig` is a union type which also consists a type called `ISequelizeStorageConfig`. And this is exactly what you are trying to solve, I assume :)

saschagehlich:
You could just copy the options from `@types/sequelize` which are way more convenient to use:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sequelize/index.d.ts#L5528

RobinBuschmann:

PR #508, opened 2019-01-10
Add `save` to IAssociationActionOptions

### LINKED ISSUE #244: Model.$set definitions conflicts with implementation

mmiszy:
Fixes #244 #493

Related documenation: http://docs.sequelizejs.com/class/lib/associations/belongs-to.js~BelongsTo.html#instance-method-set

### DIFF
--- a/lib/interfaces/IAssociationActionOptions.ts
+++ b/lib/interfaces/IAssociationActionOptions.ts
@@ -2,4 +2,5 @@
 export interface IAssociationActionOptions {
   through?: any;
   transaction?: any;
+  save?: boolean;
 }


RobinBuschmann:
👍

PR #565, opened 2019-03-08
Execute npm ci in Travis

Raul6469:
Hi 👋 
This will run `npm ci` in Travis, it ensures that the environment is reproductible. [Here is a blog post about it's benefits](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
   - npm i -g npm@5
 
 install:
-  - npm install
+  - npm ci
 
 script:
   - npm run lint


RobinBuschmann:
Hey @Raul6469 , sry for the late reply: Thanks for contributing and pointing that out. `npm ci` doesn't seem to be significantly faster than `npm i` in this repo - At least compared to build using `npm i`. Do you have a clue why? Thank you :)

Raul6469:
No, no idea :/ But at least it is designed to ensure the environment reproducibility, meaning it fails if package-lock would have been changed on install. Besides that, I agree that it doesn't make that much difference

PR #574, opened 2019-03-21
add browser support and config

Toxicable:
Adds support for blanking out all the decorator information when this package is consumed with a packager such as webpack so models defined with them can be consumed by a browser.

Im not certain if this exact config works, we have an internal hacked up version to achivee the same result


RobinBuschmann:
Hey @Toxicable thanks for contributing. Does this only work with webpack? I'm not quite sure if this should be implemented with sequelize-typescript. Probably a webpack-plugin for sequelize-typescript is more suitable. What do you think?

Toxicable:
It does definitely works with webpack. And it works in roll-up with a plug-in, not sure about other bundlers.
I think this is the best place for it, since it would be hard to process the sequalize-typescript files on the fly from a plug-in.
I got the inspiration from typeorm since they have a similar approach

Toxicable:
Just checked this and it works with the Parcel aswell, that covers all the big bundlers for  this feature.

RobinBuschmann:
I'm still not convinced. This is very specific, I'm not very familiar with this. So I don't think i'm able to maintain this. It also refers to static directories (`annotations`) - for example for `sequelize-typescript@1` the directory structure is completely different - and we would need tests for that. This is why I would prefer a separate plugin. I would definitely support you with this.

Toxicable:
@RobinBuschmann 
Doing it this way is much easier than using any sort of pugin api.
If we made plugins then we'd need a plugin for every different bundler which would be much harder to maintain. 
This is a much simplier concept (source file replacements), than learner the API of each of the bundlers.
If you have an issue with the `annotations` directory method then we could easily just commit the files output from the script I made, the reason I did it like that is just ot reudce the sice of the PR.

RobinBuschmann:
@Toxicable Can you provide more information about what really happens there? For example why is there a `noop` function needed? Thank you

Toxicable:
@RobinBuschmann Sure
That noop function is becasue the decorators need to return a function, if they don't itll throw a runtime error.
By having them all return the same function it makes it easier for tree shakers to remove it later on.

When bundling, if a bundler comes accross a statement like 
```
import { Colunm } from 'sequelize-typescript'
```
The bundler will look in `node_modules` for `sequelize-typescript` where it will look at the `package.json`, if there's a `browser` entry it will use mapping to find the `index.js` which gives the bundler a file path to resolve the `Colunm` symbol with.

for deep imports such as
```
import { Model } from 'sequelize-typescript/models/Model'
```
The bundler will follow a simlar process, except it looks up the deep path like so: `models/Model.js` in the browser mapping.
That's why we need those deep path mappings in the `package.json`, incase a user does a deep import

RobinBuschmann:
@Toxicable I don't get the point. Why doesn't the bundler load the actual files? Or why we don't want to let the bundler load the original files? Why do we need to provide noop-functions - meaning how does it help to provide references that does nothing during runtime?

Toxicable:
@RobinBuschmann The bundler cannot load the actual files since the actual files reference `sequelize` which would have to be bundled aswell, then sequelize references modules like `fs` and `path` which are not bundlable or accessible in the browser, it's not possible to bundle `sequelize-typescript` without these changes for the browser
Even if it was possible, `sequelize` is a large lib which would increase the bundle size significantly.

The point of the noop functions is becasue Decorators are runtime, therefore when the browser runs your code that has decorators it must execute them, if you don't return a function from a decorator then the browser will throw the error `Uncaught TypeError: undefined is not a function`.

RobinBuschmann:
Thanks for explaining. But I still don't fully understand it. Can you provide an example repo? I don't understand why someone should use sequelize-typescript without sequelize, since sequelize-typescript wouldn't work without it. Same for the decorators, if they don't apply their functionality, the whole library sequelize-typescript wouldn't make sense. What would be the use case of preparing sequelize-typescript for the browser like you described it?

Toxicable:
@RobinBuschmann Oh sorry, I see what you mean, I must have skipped over the whole reason for anyone wanting this feature.
From the top; If you are making an application with a TS Nodejs backend and browser TS then you'd want to share DB models between them.
Currently the approach we take is to write an interface to be shared with Server and Client, then write our models for the client to implement those interfaces.
This works but it means you'd effectively double defining your models.
eg:
```
interface Person { // to be used by browser and server
  id: number
  name: string
```
```
@Tabel
interface PersonModel implements Person {
  @Colunm
  id: number;
  @Colunm
  name: string;
}
```

The solution here is to make a single model definition in such a way where Server and Client can share the same definition, this is what is enabled by this PR. You write your models with `sequelize-typescript` decorators, wen you use them on the server they are 100% functional and able to access the DB, but when you use them on the browser they are "inert", simply TS definitions and empty classes.
This is a huge advantage since it allows you to never worry about your interfaces and models being out of sync and just makes the whole front -> back flow so much nicer to use

Toxicable:
@RobinBuschmann Do you need any more info from me on how we can support `sequelize-typescript` models in the browser?

RobinBuschmann:
@Toxicable Ah sry, yes, I've got another question and thank you very much for your explanation. Now I've seen your point.
Here my question: What about bundling an application for server side usage? For example if one need to prepare their application for `Amazon Lambdas` or `Google Firebase` - Here one would also use webpack or something similar. In that case, the sequelize-typescript functionality shouldn't be noop'd. How to overcome this with your approach?

Toxicable:
@RobinBuschmann I don't see in what situation you'd want to bundle "working" models into your frontend application, even if it is running in nodejs.
The same applies with any SSR rendered app, those models should only be there for their types not for any runtime functions. At runtime you should be hitting your api which can use the models to access the db.

publicvirtualvoid:
This feature is hugely valuable if your front end application is written in Typescript. I'm currently having to use this branch to avoid duplication on my models.

RobinBuschmann:
@toxicable I didn’t mean to use bundling for the frontend but for the backend like for AWS lambdas. If we would integrate these dummy/noop setup, does this also affect bundling for backend? Or is this leveraged by the keyword „browser“ or will webpack use the paths listed under „browser“ in the package.json by default?

Toxicable:
@RobinBuschmann Ah right, yeah this will only affect builds where browser is the target, so backend wont be effected

Toxicable:
@RobinBuschmann Do ya need anymore info on this feature?

RobinBuschmann:
@Toxicable I'm really sorry, that this take so long. Thank you for your patience! I still have one question: Since the `browser/lib/**/*`-files aren't part of the repo. When do they get generated?

Toxicable:
@RobinBuschmann Sorry for the late reply.
The script i've got in this MR will generate those files, which would be done as part of the deployment process.
If you prefer we could generate those files once then just commit those and update them with the script maybe.
or leave updates to be manual possibly.
There's a few different ways we could handle this, i'll put that to what you prefer.

RobinBuschmann:
@Toxicable no problem :) but who is triggers the generation? I see no npm script or something similar

Toxicable:
@RobinBuschmann I need to lean on you a bit for where that would be best placed, I assume tacking it onto the end of your `build` npm script would be best, do  you agree?

paqman:
Exactly what I was looking for ! Would be cool to have this merged

RobinBuschmann:
@Toxicable yes, that makes sense:
```json
{
...
"generate-browser": "node generate-browser.js",
"build": "tsc --project lib/models/v4 && tsc && npm run generate-browser",
}
```

Toxicable:
@RobinBuschmann Done, also added windows build support since my code assumed posix paths

RobinBuschmann:
👍 I wait for the build getting green and after that I'm going to publish: `sequelize-typescript@0.6.12-beta.0`

Toxicable:
Awesome! will update our repository once you cut a release.

RobinBuschmann:
@Toxicable There is still an issue with how the `browser` property within the package.json is updated. The order of the listed file paths changes on every run. Since this step runs before a package is published the working directory isn't clean anymore and the actual publish step fails: `Git working directory not clean.` Can check this?

Toxicable:
Ah bugger, right.
Thats an easy one to fix, just need to sort the list before modifying the package.json

Toxicable:
https://github.com/RobinBuschmann/sequelize-typescript/pull/627 sorted here

RobinBuschmann:
So, here we go: https://github.com/RobinBuschmann/sequelize-typescript/releases/tag/v0.6.12-beta.0

Toxicable:
Looks like I forgot that .d.ts files need to be included along with the browser js files.
Will make a patch for it

RobinBuschmann:
Just released a new beta version: https://github.com/RobinBuschmann/sequelize-typescript/releases/tag/v0.6.12-beta.1

Toxicable:
Ok, so turns out we don't need the .d.ts files in the browser.
It was the paths in the package.json, they need to be prefixed with `./`
Sending another patch.

Sorry, it's been a bit difficult to test this locally.
I have been able to run a build with this next change though, so we're there!

Toxicable:
@RobinBuschmann All working well now btw

RobinBuschmann:
Good to know :)

oscarcalvo:
This is an amazing contribution, since to share code between server/front is one of the best features of javascript stack and with interfaces at the end you are duplicating things. Thanks a lot.

I have a question @Toxicable , whats about add functions to models? I have played with the beta and with a model like

```
@Table
export class Person extends Model<Person>{
  @Column
  id: number;
  @Column
  name: string;

public validateName():boolean{
return false;
}
}
```

browser throws an exception when you use the function validateName: "TypeError: person.validateName is not a function"

is this not allowed with this new approach?

Toxicable:
@oscarcalvo Yeah so that's sort of where this is a bit weird.
While you might declare an object like 
```
var person: Person =  {name: 'foo', id: 1};
```
It's not actually an instance of the class, since class methods aren't serializable over json, meaning the above example is what the class would look like on the client.
So you have 2 options: 
1. Don't put methods in your classes
2. make a construtor take the json object
Something like this as an example
```
class Person {
  constructor(p: Person) {
    Object.entries(p).forEach(([key, value]) => this[key] = value);
  }
}
let person = new Person(await this.http.get('/person/1'))
```
I personally prefer option 1. since it's harder to mess things up if you forget to make the class instance.

oscarcalvo:
Thanks @Toxicable, I understand and even we can use Object.assign to create a new instance with methods.

Thing is: I cant use the method on browser, its not about to get the instance from server and serialize/deserialize.
if you, only using the browser, write
```
let person=new Person();
p.name="myName";
person.validateName();
```

its ok to set the name property, but browser fails telling that validateName is not a function. 

I tried to look at the source files but I dont have enough knowledge about generators. Looks like validateName is there but I dont know why it fails when calling.

oscarcalvo:
Hi @Toxicable Could you tell me something about that? I just to know if it is the default behaviour or is a bug. Thanks

Toxicable:
@oscarcalvo your code there looks fine to me, I don't know whhats going on.
One thing to note, if the function `validateName` is comming from `Model` from `sequelize-typescript` then it wont exist in the browser since `Model` is actually `class Model{}` when targetting browsers.
But if it's a method that you define in your own code, then there's nothing different between that and using pure js classes and no `sequelize-typescript`

oscarcalvo:
@Toxicable I have created a basic example in https://github.com/oscarcalvo/sequelize-typescript-frontend-model-test.git and added a model called Person. Also added a "Car" class to check that in that case all work well.

if you make "npm install" and "npm start" (note, not npm run start), and go to localhost:3000
 you will see this error in the console

Captura

Toxicable:
@oscarcalvo so I don't understand why, by replace `@Table` with `@Table({})` and it fixes the issue. I don't see how it's and issue with `sequelize-typescript`'s browser implementation, but for whatever reason this fixes it in my testing.

oscarcalvo:
I have added you to the repo so you can push because your change doesnt work for me. now it shows
TypeError: Class constructor Model cannot be invoked without 'new'

Anyway I think it needs to be fixed so to change decorator is not needed or at least well explained. My knowledge about that is very limited so I cant help there, maybe @RobinBuschmann can apport here.

Captura2

Toxicable:
That would be because you're targettign es5, sequelize-typescript requires es2015 minimum

oscarcalvo:
@Toxicable you are rigth, it was due target value. 

Anyway the problem is there still, I think  the issue with the requirement of to use @Table({}) instead of @Table needs to be fixed

PR #584, opened 2019-03-28
Update IIncludeOptions.ts

FlorianTopf:
**See implementation:** https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1643


RobinBuschmann:
Thanks for implementing

PR #595, opened 2019-04-04
Align @Column type parameter to match the Sequelize

### LINKED ISSUE #594: Support for Array types
If you define a column like:
```ts
@Column(DataType.ARRAY(DataType.STRING))
myCol: string[];
```

You'll get the error saying: 
```
error TS2559: Type 'ArrayDataType<AbstractDataTypeConstructor>' has no properties in common with type 'Partial<ModelAttributeColumnOptions>'.
```

devoto13:
See #594 for an example. This PR allows to pass any type allowed by the Sequelize to `@Column` annotation.

1. Sequelize seems to allow `string` as a type (e.g. `@Column('STRING')`), but there is a test, which fails in this case. Shall I also allow this and change test or keep it as it is now?
2. I'm not sure how to add a test for the `@Column(ARRAY(STRING))` fix as tests seems to run against Sqlite and it does not support ARRAY type. Any suggestions?

Fixes #594

### DIFF
--- a/src/model/column/column.ts
+++ b/src/model/column/column.ts
@@ -1,10 +1,10 @@
-import {ModelAttributeColumnOptions, DataTypeAbstract} from 'sequelize';
+import {ModelAttributeColumnOptions, DataType} from 'sequelize';
 
 import {addAttribute} from './attribute-service';
 import {isDataType} from "../../sequelize/data-type/data-type-service";
 import {getSequelizeTypeByDesignType} from "../shared/model-service";
 
-export function Column(dataType: DataTypeAbstract): Function;
+export function Column(dataType: DataType): Function;
 export function Column(options: Partial<ModelAttributeColumnOptions>): Function;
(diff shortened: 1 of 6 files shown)


RobinBuschmann:
@devoto13 Thanks for fixing this. Regarding 1. what test fails here? According to travis all tests get passed. Regarding 2., we could add a `types` folder to the `test` dir and just compile the files in there in order to test the typings. Since this is only think we want to test here: The typings. What do you think?

devoto13:
1. [This one](https://github.com/RobinBuschmann/sequelize-typescript/blob/1.0.0/test/specs/utils/data-type.spec.ts#L31). Since string is allowed `'abc'` will be considered a data type as well. See the second commit in the PR for the change I'm talking about.

2. Good idea! Added a spec, which will fail compilation if does not compile.

RobinBuschmann:
Good Job!

PR #619, opened 2019-05-14
Moving sequelize and node to peerDependencies

dancrumb:
Moving sequelize and node to peerDependencies

This makes it a lot easier to fix/patch type issues that sequelize-typescript depends on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,9 +57,11 @@
     "@types/chai-datetime": "0.0.30",
     "@types/lodash": "4.14.109",
     "@types/mocha": "2.2.39",
+    "@types/node": "11.12.2"
     "@types/prettyjson": "0.0.28",
     "@types/sinon": "1.16.35",
     "@types/sinon-chai": "2.7.27",
+    "@types/sequelize": "4.27.48",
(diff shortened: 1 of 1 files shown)


RobinBuschmann:
Hey @dancrumb thanks for contributing. To really give the user of sequelize-typescript the ability to install their own typings, the typings dependencies also have to be less strict. Nevertheless this would mean a major change, which is already done with `sequelize-typescript@1.0.0-beta..`: https://github.com/RobinBuschmann/sequelize-typescript/blob/1.0.0/package.json#L73

dancrumb:
That's fair.

I wonder - since Promises have been supported since 0.12 (https://node.green/#ES2015-built-ins-Promise) and the recommended level these days is 10.15, perhaps it's time to drop the references to bluebird and rely on native Promises?

RobinBuschmann:
@dancrumb that's up to sequelize whether or not bluebird or native promises are used. I think sequelize uses features that native promises do not offer - that's why they don't switch. But this is nothing we from sequelize-typescript can decide. But thanks for your hint :)

Issue #613, opened 2019-05-05
Group should be included in include if seperate:true is specified


chaitanya11:
If seperate: true is specified, user should be able to do group operation, because it will be executed as separate query.
related issue in [Sequlize](https://github.com/sequelize/sequelize/issues/1640#issuecomment-40706881).

[stack overflow question](https://stackoverflow.com/questions/37339708/sequelize-group-by-aggregating-only-on-main-table-instead-of-the-complete-joined)

RobinBuschmann:
Hey @chaitanya11, thanks for bringing this up. Can you fill out the [issue template](https://github.com/RobinBuschmann/sequelize-typescript/blob/master/ISSUE_TEMPLATE.md) please?  I don't quite understand your issue. Are you expecting [this](https://github.com/sequelize/sequelize/issues/1640#issuecomment-40706881) to work with sequelize-typescript while it doesn't?

chaitanya11:
```
QUESTION_MAIN.findAll({
attributes:['QUESTION_ID', 'POSTER_I',
  ['( 6371 * acos( ' 
                  + 'cos( radians('+qstnFeedRequest.qstnLocLat+') ) ' 
                  + '* cos( radians( QSTN_LOC_LAT ) ) '
                  + '* cos( radians( QSTN_LOC_LONG ) - radians('+ qstnFeedRequest.qstnLocLong+') ) '
                  + '+ sin( radians('+qstnFeedRequest.qstnLocLat+') ) '
                  + '* sin( radians( QSTN_LOC_LAT ) ) ) '
  + ')', 'DISTANCE'
  ]
],
include: [
  { model: USER_ATTRIB, 
    attributes:['USER_NAME']
  },
  { model: QSTN_CATG, 
    attributes: [['CATG_I', 'QSTN_CATG_I']],
    where: qstnCatgWhereClause
  },
  { model: REPLY_MAIN, 
    attributes: ['QSTN_I', [sequelize.fn('COUNT', sequelize.col('REPLY_MAIN.QSTN_I')),   'REPLY_COUNT']], 
    where: {REPLY_STATUS: 200},
       group: ['QSTN_I'], // Should be enabled if it is a separate query
       separate: true, // separate query
    having: {'REPLY_COUNT': {$ne: null}}, 
    required: false
  }
],
having:{ 'DISTANCE' : {$lt: 5} },
where: whereClause,
limit: qstnFeedRequest.limit
})
```
[...]

PR #624, opened 2019-05-24
Adding group typings for separate query

chaitanya11:
Adding fix for issue #613, Added required typings to avoid compilation error.


RobinBuschmann:
@chaitanya11  Thanks for implementing 👍

PR #632, opened 2019-06-02
make index browser file path relative aswell

Toxicable:


Toxicable:
I missed the index file in this change here:
https://github.com/RobinBuschmann/sequelize-typescript/commit/244b16cfaa19dbf2888dad551720d7e3756cfa3c

This is the last thing needed to close to support browser

Toxicable:
@RobinBuschmann ping would be great to merge this in

PR #633, opened 2019-06-04
Fix IFindOptions typings

lukemuller:
Similar to https://github.com/RobinBuschmann/sequelize-typescript/issues/226


RobinBuschmann:
Thanks for fixing

Issue #668, opened 2019-08-23
[feat] Add browser support back in


Toxicable:
It would appear in the 1.0.0 merge the generate-browser.js script was skipped over.
This was the latest version: https://github.com/RobinBuschmann/sequelize-typescript/blob/a05f4fcaa619ab04455ca87a39dacbeb8447a2c1/generate-browser.js

Would be great to have browser support back in for 1.0.0

RobinBuschmann:
Hey @Toxicable, thanks for bringing this up. In Version 1.0.0 the folder structure completely changed compared to 0.6.X. So that much more manual work is needed to bring this feature back alive.

Are you using real implementations (e.g. class functions) or just the types of models in your client? I'm asking because if the latter, maybe a mapped type would be a better solution?

RobinBuschmann:
And sorry for the late reply

Toxicable:
It's all good.
I came up with a far more simplified version here: https://github.com/RobinBuschmann/sequelize-typescript/pull/673

As for how we use them, we only use properties, we don't define any functions in out Models.
We actually go a bit further than athat and noop out all the Sequalize types on Model like this:

```
class NoopClass {}
const BaseModel: typeof NoopClass = Model as any;

@Table
export class Person extends BaseModel {
  @Colunm  
  name: string
}
```
This is so you don't get methods like `findById()`, which wont work in the browser.
Instead on the server we have a repository wrapped which use to access the DB

RobinBuschmann:
I'm just wondering, that the bundler tries to bundle the models at all, since you only use them as types. Actually they should be omitted in general.

```ts
import {User} from './models/user';
import _ from 'lodash';

const user: User = _.assign({}, {...});
```
should become something like:
```js
var _ = require('lodash');

var user = _.assign({}, {...});
```

Nevertheless, I'm totally fine with your solution (#673). Thanks for implementing it!

PR #673, opened 2019-09-07
add browser support back in

Toxicable:
This is a much more simple implementation than the previous one. 

Just 1 file authored in TS, so updates are simple and we get type checking.

The only downside is it dosen't support deep imports, but the 1.0 release broke those and I wouldn't consider it part of the public api anyway.

Closes https://github.com/RobinBuschmann/sequelize-typescript/issues/668


RobinBuschmann:
@Toxicable released under `sequelize-typescript@1.1.0-beta.0`. Can you confirm that it works as intended?

Toxicable:
@RobinBuschmann Yup, looks like it's working to me!

PR #733, opened 2020-01-16
Update README.md - Fix broken links

przemwierzbicki:
As `sequelize` documentation changes, so should the links in the docs for `sequelize-typescript`.

This little contribution from a concerned Junior Dev (who felt the weight of this problem on his own skin as he tried to navigate the original docs himself, alone against the broken links) makes sure it is so :)


RobinBuschmann:
Thank you for fixing this!

PR #736, opened 2020-02-01
RE-fixing a bug I fixed already.

TimMensch:
Issue 446 requires this fix. I submitted a fix for it a long time ago, but apparently the fix was reverted eventually?

So I'm submitting it again with more documentation in hopes that it will _stay fixed_ this time.

Ideally Sequelize itself would be fixed, but there are a dozen or more _closed_ references_ to this bug in their issues and many dozen PRs being ignored, so I don't have high hopes it will be fixed.


RobinBuschmann:
Hey @TimMensch, thanks for bringing this up again. Do you know why this isn't fixed in sequelize either? Since `include` is explicitly deleted, it seems to me that there is a specific reason for that. This line for example depends on `options.include` is undefined, rather than an empty array: https://github.com/sequelize/sequelize/blob/0fb4bb11f3771a4b229cc054ee56818c10b5b395/lib/dialects/abstract/query-generator.js#L1161

TimMensch:
I submitted a parallel PR to Sequelize itself:

https://github.com/sequelize/sequelize/pull/11893

...and then tagged it in this bug that showed up the next day:

https://github.com/sequelize/sequelize/issues/11894

It's been sitting there, ignored, for more than a week now. Sequelize has nearly 60 open PRs, some dating back _months_. This particular bug was noted as far back as 2014, when the response was "don't call count without setting `col`":

https://github.com/sequelize/sequelize/issues/2713

So the bug has been in Sequelize for at least 6 years, maybe longer. Sequelize just generates invalid code for `findAndCountAll` without the `col` value set, but with `distinct` true. I have a sense that the developers are spending all of their time and energy working on v6, because of course that's the fun thing to do. Oh, and v6 doesn't fix the bug _either_.

I'm using `sequelize-typescript` with `feathers-sequelize`, which calls `findAndCountAll` without the `col` value set, but instead with the workaround of setting `include` to `[]` so that Sequelize actually generates valid SQL. You can see their workaround here, along with a note that Sequelize has a bug:

https://github.com/feathersjs-ecosystem/feathers-sequelize/blob/master/lib/index.js#L140

Maybe `feathers-sequelize` needs to set `col` directly? I don't know. I'm setting it now in a Feathers hook for all queries, and I'm hoping that doesn't break anything, though honestly it might. As it stood, `feathers-sequelize` + `sequelize-typescript` simply doesn't work out of the box; you try any query at all and it throws a completely opaque error.

Since the Sequelize folks seem to be practically ignoring PRs and issues (750 open issues?!), and they've already ignored this bug for years despite it being a critical illegal code generation bug, I despair of it being fixed there. I could try to request a fix in `feathers-sequelize`, to add `col=primaryKey` to every `findAndCountAll` call where `distinct` is true, but the right answer might actually be to count distinct _rows_ (i.e., list all the columns individually in the distinct call).

The line you're linking in the Sequelize source means, if I'm interpreting it correctly, that the queries it generates will be more verbose _if someone passes in_ `includes=[]`. I have to say "oh well, don't do that, if you care!" The problem is, _without_ this hack (or another workaround), Sequelize generates illegal SQL and crashes the query.

And if you redirect me to the `feathers-sequelize` project, and I can't convince them to care about `sequelize-typescript` breaking _their_ otherwise valid workaround, then I'm completely stuck with no way to actually make this work without forking either your project or theirs.

Divlo:
Is this bug has been fixed since then ? @TimMensch

TimMensch:
Actually I think that Sequelize upstream accepted my PR, and it therefore may be fixed there.

But I'm moving away from using Sequelize entirely, so I'm not prepared to do the work required to track this down. I'll just close it.

PR #774, opened 2020-04-23
Allow $set null (remove association)

Antiavanti:
Referring to the documentation, use of `null` for relation `$set()` should be allowed.
https://sequelize.org/v5/manual/associations.html#associating-objects
```
// project is associated with task1 and task2
task2.setProject(null).then(() => {
  // and it's gone
})
```


Divlo (approved):

RobinBuschmann (approved):

engmsilva:
I'm not getting a way to remove an association between these two models below:

```
@Table
class Ticket extends Model {
  @PrimaryKey
  @AutoIncrement
  @Column
  id: number;

  @ForeignKey(() => Queue)
  @Column
  queueId: number;

  @BelongsTo(() => Queue)
  queue: Queue;
  
}
```

```
@Table
class Queue extends Model {
  @PrimaryKey
  @AutoIncrement
  @Column
  id: number;
}
```


To associate a Queue to a Ticket I do the following:

```
const queueId = 1;
await Ticket.findByPk(ticket.id).then(async ticket => {
          await ticket.update({
            queueId
          });
ticket.reload();
  });
```

Any ideas on how to remove this association?

PR #831, opened 2020-08-27
Update association-get-options.ts

oktapodia:
We can also specify multiple scopes through that option.

```javascript
// house model with a persons associations
house.$get('persons', { scope: ['includeChildrenStat', 'includeElderStat'] })

```


Divlo (changes requested):
Hey, thank you for your contribution! @oktapodia 
Could you have a link to the `sequelize` docs about this feature ?

PR #865, opened 2020-12-08
README update for sequelize v6

lukashroch:
Update readme

- how to upgrade to sequelize v6 & sequelize-typescript v2
- outline breaking changes
- sequelize-typescript v1 with v5

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,10 @@
 [![Dependency Status](https://img.shields.io/david/RobinBuschmann/sequelize-typescript.svg)](https://www.npmjs.com/package/sequelize-typescript)
 
 # sequelize-typescript
-Decorators and some other features for sequelize (v5).
+Decorators and some other features for sequelize (v6).
 
  - [Installation](#installation)
+ - [Upgrade to `sequelize-typescript@2`](#upgrade-to-sequelize-typescript2)
  - [Upgrade to `sequelize-typescript@1`](#upgrade-to-sequelize-typescript1)
  - [Model Definition](#model-definition)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,29 +38,96 @@ Decorators and some other features for sequelize (v5).
...
+import { Table, Model } from 'sequelize-typescript';
+
+interface PersonAttributes {
+  id: number;
+  name: string;
+}
+
+interface PersonCreationAttributes {
+  name: string;
+}
```
intellix, line 125:
```suggestion
interface PersonCreationAttributes extends Omit<PersonAttributes, 'id'> {
}
```

Divlo (approved):

README.md
lukashroch: Thanks @intellix, added with corresponding import statement. Good idea to keep it consistent with sequelize official TS guide.
...though anyone can freely to construct the interfaces however feels appropriate.

Divlo (approved):

DaPfeifa:
I hope this can be merged soon.
Otherwise you always run into problems when migrating to Sequelize v6.

lukashroch:
I've squashed the changes and added a note about current state of implementation, so it's more clear what currently works.

- less strict sequelize V6 works quite fine - this is equivalent to what we had so far
- more strict sequelize introduced in v6.2+ is not yet implemented, will need more work

@RobinBuschmann if this could be merged, it would be much appreciated. I should have done the README update along with the code change, so users knew current state and what needs to be done. Thanks!

PR #889, opened 2021-01-21
:sparkles: add virtual field to README

### LINKED ISSUE #832: Doc request: VIRTUAL field and it's getter setter
**I'm submitting a ...** 

Doc request

I know how to write a Virtual field in sequelize, but not clear for `sequelize-typescript` class model
All I can find are two issues:
https://github.com/RobinBuschmann/sequelize-typescript/issues/663
https://github.com/RobinBuschmann/sequelize-typescript/issues/701

But still not very clear about:
- Do I need to put the dependent fields in type? (like `DataType.VIRTUAL(DataType.STRING, ['first_name', 'last_name'])`)
- How do I write a setter along with getter

So, maybe write some small examples for VIRTUAL fields in doc would help people

goalia:
Add documentation about virtual field. I found 3 issues who mentioned this function. ([Sequelize documentation](https://sequelize.org/master/manual/getters-setters-virtuals.html)
This can close #832

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Decorators and some other features for sequelize (v6).
  - [Model validation](#model-validation)
  - [Scopes](#scopes)
  - [Hooks](#hooks)
+ - [Virtual field](#virtual-field)
  - [Why `() => Model`?](#user-content-why---model)
  - [Recommendations and limitations](#recommendations-and-limitations)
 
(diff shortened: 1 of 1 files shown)


goalia:
@Divlo @lukashroch @RobinBuschmann @lukashroch :)

README.md
```diff
@@ -921,6 +922,34 @@ export class Person extends Model {
...
+@Table({tableName: 'users'})
+class User extends Model {
+ 
+  @Column(DataType.VIRTUAL)
+  get secretToken() {
+    return this.getDataValue('secretToken');
+  }
+  set secretToken(value: string){
+    this.setDataValue('secretToken', value)
+  }
```
Divlo, line 943: We could use the same example as the `sequelize` documentation.

README.md
```diff
@@ -921,6 +922,34 @@ export class Person extends Model {
...
+}
+```
+After that you can populate the field like so:
+```typescript
+const user: User = await User.findOne({where: {id});
+user.secretToken = secretToken;
+return user;
```
Divlo, line 950: Why there is this return, we're not inside a function. 🤔

Divlo (changes requested):
See my comments. Thanks for your contribution! @goalia

PR #868, opened 2020-12-13
Fix typeof Model errors throughout by using typeof Model generics

intellix:
Hopefully addresses https://github.com/RobinBuschmann/sequelize-typescript/issues/835

This is essentially me just applying it in the places that worked and then re-applying it throughout until typescript errors disappeared and it compiled/built. 

It builds and all tests pass. When I include it in my own project I go from 457 errors to 107 and I don't see my original problems as in issue #835. 

```
  493 passing (27s)
  6 pending


=============================== Coverage summary ===============================
Statements   : 97.95% ( 1098/1121 )
Branches     : 89.46% ( 280/313 )
Functions    : 97.27% ( 249/256 )
Lines        : 97.99% ( 1075/1097 )
================================================================================
```

I'm still getting a problem with Includeable from Sequelize as it expects `typeof Model`. The new ModelType in this PR would fix it, but how do you patch existing types?
For example:
```
PaymentCard.findOne({
  transaction,
  include: [BillingAddress], <-------------
  where: { active: true },
}),
```

Gives: 

```
No overload matches this call.
  Overload 1 of 2, '(this: ModelStatic<PaymentCard>, options: NonNullFindOptions<PaymentCardAttrs>): Promise<PaymentCard>', gave the following error.
    Type 'typeof BillingAddress' is not assignable to type 'Includeable'.
      Type 'typeof BillingAddress' is not assignable to type 'typeof Model'.
        Types of construct signatures are incompatible.
          Type 'new (values?: BillingAddressCreateAttrs, options?: BuildOptions) => BillingAddress' is not assignable to type 'new <TModelAttributes extends {} = any, TCreationAttributes extends {} = TModelAttributes>(values?: TCreationAttributes, options?: BuildOptions) => Model<TModelAttributes, TCreationAttributes>'.
            Types of parameters 'values' and 'values' are incompatible.
              Type 'TCreationAttributes' is not assignable to type 'BillingAddressCreateAttrs'.
                Type '{}' is missing the following properties from type 'BillingAddressCreateAttrs': paymentCardId, name, lastName, address1, and 6 more.
  Overload 2 of 2, '(this: ModelStatic<PaymentCard>, options?: FindOptions<PaymentCardAttrs>): Promise<PaymentCard>', gave the following error.
    Type 'typeof BillingAddress' is not assignable to type 'Includeable'.
      Type 'typeof BillingAddress' is not assignable to type 'typeof Model'.ts(2769)
```

[...]

PR #900, opened 2021-02-06
Fix typeof Model errors throughout by using typeof Model generics

KapitanOczywisty:
Update of #868 with fixed lint errors. https://github.com/RobinBuschmann/sequelize-typescript/pull/868#issuecomment-774444691

cc: @Divlo


src/associations/belongs-to-many/belongs-to-many-association.ts
```diff
@@ -8,28 +8,29 @@ import {ModelClassGetter} from "../../model/shared/model-class-getter";
...
     const options: OriginBelongsToManyOptions = {...this.options as any};
     const associatedClass = this.getAssociatedClass();
     const throughOptions = this.getThroughOptions(sequelize);
 
     const throughModel = typeof throughOptions === 'object' && typeof throughOptions.model !== "string" ? throughOptions.model : undefined;
     options.through = throughOptions;
-    options.foreignKey = getForeignKeyOptions(model, throughModel, this.options.foreignKey);
-    options.otherKey = getForeignKeyOptions(associatedClass, throughModel, this.options.otherKey);
+    options.foreignKey = getForeignKeyOptions(model, throughModel as any, this.options.foreignKey);
+    options.otherKey = getForeignKeyOptions(associatedClass, throughModel as any, this.options.otherKey);
```
Divlo, line 33: Is there is no way to avoid this `as any` type ?

Divlo (changes requested):
Thanks for your contribution! @KapitanOczywisty 
That's great to have a stricter model defintion!

What if I don't want to provide `<TCreationAttributes, TModelAttributes>` is it still working ?
It should still works!

Also see my comment. :smile: 
We should avoid as much as possible `any` type, sometime we can't, we need to use `any`, we can't do something else.

src/associations/belongs-to-many/belongs-to-many-association.ts
KapitanOczywisty: @Divlo You can see `any` in few other places in this file, seems like this was badly designed to begin with and it's beyond scope of this PR to fix this. I'm not seeing better fix without major redesign and getting modal attributes to work is way more important. Argument is not affecting return type anyway.

KapitanOczywisty:
> What if I don't want to provide `<TCreationAttributes, TModelAttributes>` is it still working ?

In my testing both ways work now, seems like @intellix did a great job. When you don't provide attributes you have effectively `<{}, {}>`.

src/associations/belongs-to-many/belongs-to-many-association.ts
Divlo: Right, maybe we'll need to refactor this code, someday.
I'll approve for the moment.

Divlo (approved):

RobinBuschmann (changes requested):
Hey @KapitanOczywisty thanks for fixing this. Can you add some tests please?

KapitanOczywisty:
@RobinBuschmann I probably could add a couple of types-only tests. What do you want to see in these tests?

RobinBuschmann:
> @RobinBuschmann I probably could add a couple of types-only tests. What do you want to see in these tests?

The tests should validate that it's now possible to use `<TCreationAttributes, TModelAttributes>`. Yeah, checking the types should be enough 👍

KapitanOczywisty:
Sooo, I see problems near everything what is touching `through` in `BelongsToMany` I have to rewrite this a bit.

And there are also problems with `FooModel.findAll({include: BarModel})`, but this is due to sequelize itself, so leave that for them to deal with?

KapitanOczywisty:
@RobinBuschmann Fixed BelongsToMany and some testing added.

KapitanOczywisty:
For userland type-guard have to be a bit more finicky and I'm thinking about exporting this one to userland, where you need to access some static properties from `Model`. I'd like to hear some feedback on that.

Generally, `type ModelType = Properties<typeof Model>` also works fine as type-guard and is fine when you don't care about attributes (which is the case in example bellow).

```ts
type Properties<T> = {
  [P in keyof T]: T[P];
}
type ModelType<TModelAttributes = any, TCreationAttributes = TModelAttributes> = Properties<typeof Model> & (new () => Model<TModelAttributes, TCreationAttributes>);

// ---

export function SequelizeFilter(model: ModelType, filterable: string[], filter?: string): WhereOptions | undefined {
  if (!filter || filter === "") return;

  const fields = [];
  for (let prop of filterable) {
    // rawAttributes works here because of Properties<typeof Model>
    // otherwise there would be: Property 'rawAttributes' does not exist on type 'ModelType<any, any>'.
    if (model.rawAttributes?.[prop]) {
      fields.push({ [prop]: { [Op.substring]: filter } });
    }
  }

  return fields.length ? { [Op.or]: fields } : undefined;
}
```

RobinBuschmann (approved):
Sorry for the late reply: Good work 👍

Divlo (approved):

intellix:
Awesome, thanks for moving this forward :)

It seems my initial issues weren't addressed though in regards to Includeables: https://github.com/RobinBuschmann/sequelize-typescript/pull/868#issue-539045477

PR #871, opened 2020-12-23
Add automated release process and fix tslint errors

Divlo:
Hey @RobinBuschmann, hopefully everything is fine for you! :heart: 
Just wanted to know, what is the current state for adding colaborators to the project and how is going with the automated release process via GitHub as discussed in #856.

If you need some help to maintain `sequelize-typescript`, I would be happy to do so. 😃

**What changes this PR introduce ?**

- Update dependencies to the latest version so it fixes vulnerabilities
- Delete the `package-lock.json` file since packages on npm doesn't have lockfiles (I added a rule for that in `.npmrc)`.
- Add `.github/dependabot.yml`
- Fix `ts-lint` errors since the current build on `master` is failing
- Delete `.travis.yml` file and instead use GitHub Actions as it is better integrated with GitHub and they change their pricing model, see: [blog post](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)
- Create a GitHub Action to automatically release the package on npm when someone create a new release on GitHub (see `.github/workflows/npm-publish.yml)`
You will need to set a new secret `NPM_TOKEN` on the repo so it could works as it is [explained there](https://docs.github.com/en/free-pro-team@latest/actions/guides/publishing-nodejs-packages).

PR #913, opened 2021-02-14
chore(release): v2.1.0

Divlo:
**What changes this PR introduce ?**

- Delete `Cache dependencies` step in CI so hopefully it solves the issue with the CI failing (as discussed in Gitter chat)
- Downgrade `sqlite3` from `5.0.1` to `5.0.0` since `sqlite3@5.0.1` have issues. See: https://github.com/mapbox/node-sqlite3/issues/1424
- Add `husky` so everytime we commit, it verify the commit message, and run `npm run lint` also to avoid installing husky when installing the package, we use `pinst` made by the creator of `husky`. See: https://github.com/typicode/pinst
- Add `release-it` and `commitlint` so it generates the changelog automatically based on commit messages also automatically choose the version number to increase based on semver.
- Add `markdownlint`, see: https://github.com/DavidAnson/markdownlint
- Cleanup the `README`, so we only display how to use the package for the current latest version, if users want to know how to use it for older versions of TypeScript they can see in the Changelog.

**Is there anything you'd like reviewers to focus on?**

Once this PR has been merged, we could create a new release on GitHub, so it automatically release a new version on `npm` if you set correctly `NPM_TOKEN` secret as it was suggested in this PR: #871.

We should use https://github.com/zeke/semantic-pull-requests.

**Note**: Git is showing a large number of changes but in fact, it's because I installed a few new dependencies so `package-lock.json` has lot of changes.


.github/workflows/nodejs.yml
```diff
@@ -24,12 +24,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: 'Cache dependencies'
-        uses: 'actions/cache@v2.1.4'
-        with:
-          path: '**/node_modules'
-          key: ${{ runner.os }}-${{ hashFiles('**/package.json') }}
-
       - run: 'npm install'
```
lukashroch, line 27: I would use `npm ci` instead of `npm install` for ci workflow, https://docs.npmjs.com/cli/v7/commands/npm-ci

RobinBuschmann (approved):
👍

lukashroch (approved):

PR #915, opened 2021-02-15
chore(linter): replace deprecated tslint with eslint

lukashroch:
Following recent build/repo maintenance improvements, TSLint has been deprecated for a while (https://palantir.github.io/tslint/). PR swaps tslint with eslint with sensible minimal defaults to keep the code cleanly formatted.

- [X] Tests passing
- [X] Build OK


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
Divlo, line 1: I guess it's better to not use custom file extensions when it is possible, it's better to use regular/common file extensions.
For example for `.npmrc`, we can't have another file extensions (as far as I know).

But for `prettier`, we can, for example, we could use `.prettierrc.json` instead, so for IDE that don't know `prettier`, we can still have auto completion and syntax highlighting etc, as it is the case in `github` (syntax highlighting doesn't work on `.prettierc` file).

.eslintrc.js
```diff
@@ -0,0 +1,26 @@
...
+  parserOptions: {
+    ecmaVersion: 2021,
+    sourceType: 'module',
+  },
+  plugins: ['prettier', '@typescript-eslint'],
+  rules: {
+    '@typescript-eslint/ban-types': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/explicit-module-boundary-types': [
+      'error',
```
Divlo, line 20: Could you tell a little bit more of why you disabled these rules ?

package.json
```diff
@@ -6,7 +6,8 @@
     "build": "tsc",
     "test": "mocha test/**/*.spec.ts",
     "cover": "nyc mocha test/**/*.spec.ts",
-    "lint": "tslint -c tslint.json --project tsconfig.json",
+    "lint": "eslint --ext .ts src/ test/",
+    "lint:fix": "eslint --ext .ts src/ test/ --fix",
```
Divlo, line 10: Maybe we could add a new `format` npm script to run `prettier`.

Also instead to use `--ext` flag, since it's repetitive to have it twice with the same patterns, we should have it in the `.eslintrc.js` config file.
See: <https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns>

src/browser/index.ts
```diff
@@ -1,4 +1,9 @@
-function noop() { return (target: Object, propertyName: string) => {}; }
+function noop() {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Divlo, line 2: We should avoid as much as possible to use `eslint-disable-next-line` comments.
What does the function ? It seems it does nothing, maybe we don't need it anymore.

Divlo (changes requested):
Indeed, `tslint` is now deprecated and we should not use it. :+1:

In my opinion, it would be better to not have too much configuration for linter or prettier.
Maybe we could use a bigger configuration used by many companies likes `standard` or `airbnb` style.
We could look into `ts-standard`, it has great rules, and clean. Just my random thought, feel free to disagree.
The advantages, is that instead to manage rules or dependencies, we only have 1 dependency : `ts-standard`, that's all.

Although, I review your PR with `eslint` and `prettier` configuration (see my comments). :smile:

.prettierrc
lukashroch: I've worked on projects with all these type of files and never had issue with any of them. Any decent IDE can handle them accordingly. So I guess it's a matter of preference :-) I think both prittier and eslint support no extention, json and js. I like js to keep syntax consistent with code, but it's easily changable.

.eslintrc.js
lukashroch: To conform with current codebase style and how the code was written. Basically not to unnecessarily bend the code around the someone else's defaults, but make the tool work for you.

package.json
lukashroch: prettier is used as eslint's plugin, so it run together, no need for additional command.

src/browser/index.ts
lukashroch: Agreed, there are minimal occurrences in the code with these exceptions. But again, I didn't want to rewrite the codebase just because someone's rule has different opinion :-)

lukashroch:
Hi @Divlo,

thanks for reviewing. The intention was mostly to try get rid of something deprecated and pick a similar eslint style the codebase was written with with minimal disruption. This is actually a very minimal config. If you look up some open source libs, people have a very lengthy eslint configs with tons of rules turned on/off or modified rules to make their codebase look to their image :-) 

Changing it to something like `air-bnb` is easy enough, just a matter of pulling more configs and adding line to config. But that's actually a way more configuration, just hidden within the package ;-) And these predefined styles are actually way more opinionated then default rules coming in with tools.

Linting is often very opinionated, so I tried not to introduce too much opinion and keep the codebase with very little change, but to make it more consistent with already "used" style it was written with.

.prettierrc
Divlo: I don't think it's a matter of preference, and I don't see any advantages to use `.prettierc`.
> I like js to keep syntax consistent with code, but it's easily changeable.

Why if you like, js extensions, you didn't use it for the `prettier` config ?
You used it for `eslint` but not for `prettier`, it makes the code inconsistent.

src/browser/index.ts
Divlo: Right, maybe we'll need to look, what need to be done to remove this comment before merging.

Divlo:
To be honest, I didn't look the `airbnb` config, but for `ts-standard`, it doesn't at all add more configs, it's the contratry, we don't have config files with `ts-standard`, and instead to install `eslint`, `prettier` and all the `eslint` plugins as dependencies, we only need `ts-standard` nothing else (under the hood, it uses `eslint`, and all that stuff, but we don't need to configure it ourselves).
I agree that part is totally about preferences, so it's up to you, I just wanted to share that others libraries could simplify us the linting part.

Right, good point, I totally agree :+1:  :
> Linting is often very opinionated, so I tried not to introduce too much opinion and keep the codebase with very little change, but to make it more consistent with already "used" style it was written with.
> The intention was mostly to try get rid of something deprecated and pick a similar eslint style the codebase was written with with minimal disruption

.eslintrc.js
Divlo: I think, we should not personalize too much the `eslint` rules, as you said :
> Linting is often very opinionated, so I tried not to introduce too much opinion and keep the codebase with very little change, but to make it more consistent with already "used" style it was written with.

We should keep the default `eslint` rules and try to rewrite what need to be rewrite if the default rules aren't working.

.prettierrc
lukashroch: Oh god, sorry, I haven't been paying too much attention. Fixed in new commit

package.json
lukashroch: I've just been looking into this, it seems that config file options for glob patterns are only for overrides, not for top level execution... or am I missing something?

lukashroch:
Just checked the `ts-standard` and to be honest, I'm not big fan on these meta preconfigured solutions. And since it's only a preconfigured bundle, it very opinionated already just by its nature.
You're right, for your `package.json` file it's a one liner to add it. But behinds the scenes it pulls all those packages anyway + all `clutter` you might actually don't want. So in the end you end up with bigger dependency tree. On top off that, they provide their "own configuration". So in the end when you want to customize something, you brought much complexity in. I guess these are more useful for end products to have unified approach for various projects in companies, but I don't know if it a good fit for open source lib.

Just a quick comparison with one of the top eslint configs: https://www.npmtrends.com/ts-standard-vs-eslint-config-airbnb

src/browser/index.ts
lukashroch: Any ideas...? :-) It's not a bad thing to very sparingly use it (there is a reason why eslint allows disables like this ) ... If use them too often, you should rethink your ruleset. I mean e.g. in this case, I can see why it's written like this to keep the code consistent about the implementation ... even though vars are not used. If you remove them, you loose a bit of context and about what arguments are accepted. So there is always some trade-off.

Divlo:
What's up about this PR ?
I still think it is great to migrate from `tslint` to `eslint` and `prettier`. :+1:

lukashroch:
Not approved, so not much really :-)

Divlo (approved):
My comments was essentially saying that we should keep the default `eslint` rules, but other than that, everything is looking good, and as you said it is to conform with the current codebase style and how the code was written, so we don't break anything or changing the codebase too much.

I guess we could probably merge this, so we could avoid conflicts when people are doing PRs later, and it is much better to not use deprecated tools.
Approved! :+1:

PR #1070, opened 2021-08-06
fix: use custom decorator on column have a property descriptor

caolvchong:
we add custom decorator(`@CustomField`) for a column to change it to be an accessor. for example:

```typescript
function CustomField(target: any, key: string): void {
  Object.defineProperty(target, key, {
    configurable: true,
    enumerable: true,
    get() {
      const value = this.getDataValue(key);
      return `prefix/${value}`;
    },
    set(value: string): void {
      const _value = value?.replace(/^prefix/, '');
      this.setDataValue(key, _value);
    }
  });
}
```

```typescript
@Table
export class ModelA extends Model {
  @Column
  name: string;

  @Column({
    type: DataType.VIRTUAL
  })
  @CustomField
  bio: string;
}
```

```typescript
@Table
export class ModelB extends Model {
  @PrimaryKey
  @AutoIncrement
  @Column
  id: number;

  @Column({
    type: DataType.VIRTUAL
  })
  @CustomField
  bio: string;
}
```

the `@Column` decorator will lose propertyDescriptor(because class field decorator does not hava descriptor)


caolvchong:
@RobinBuschmann have a look pls

jagged3dge (approved):

jagged3dge:
Apologies for the errant review. 

Tagging @RobinBuschmann for review

Divlo (approved):
LGTM!
Thanks for your contribution! :+1:

Issue #725, opened 2020-01-05
createIndexDecorator has problems with "underscored" or "field" params


piotrmoszkowicz:
**Versions**
- sequelize: 5.21.3
- sequelize-typescript: 1.1.0
- typescript 3.7.4

**I'm submitting a ...** 
[x] bug report
[ ] feature request

**Actual behavior:**
I've tried using new feature - `createIndexDecorator`, but it doesn't take into account the `underscored` table param (`field` param on the column doesn't work either) - therefor it errors out on database level, because there are no such fields existing.

**Expected behavior:**
Successful creation of index.

**Steps to reproduce:**
Code is below.

**Related code:**

```ts
const OwnedItemUniqueIndex = createIndexDecorator({type: "UNIQUE"});

@Table({
  tableName: "game_owned_items",
  underscored: true
})
export default class OwnedItem extends Model<OwnedItem> {
  @PrimaryKey
  @AutoIncrement
  @Column
  public id: number;

  @OwnedItemUniqueIndex
  @Column
  public ownerId: number;

  @OwnedItemUniqueIndex
  @Column
  public itemId: number;
}
```

piotrmoszkowicz:
"Temporary fix"

```ts
const OwnedItemUniqueIndex = createIndexDecorator({ type: "UNIQUE" });

@Table({
  tableName: "game_owned_items",
  underscored: true
})
export default class OwnedItem extends Model<OwnedItem> {
  @PrimaryKey
  @AutoIncrement
  @Column
  public id: number;

  // @ts-ignore
  @OwnedItemUniqueIndex({ name: "owner_id"})
  @ForeignKey(() => Soldier)
  @Column
  public ownerId: number;

  // @ts-ignore
  @OwnedItemUniqueIndex(({ name: "item_id" }))
  @ForeignKey(() => Item)
  @Column
  public itemId: number;
}
```

Works with PostgreSQL.

RobinBuschmann:
Hey @piotrmoszkowicz, thanks for reporting.

jessemyers:
I have a similar problem where:

ErikMartensson:
I also have this issue, and as jessemyers have pointed out, the problem is directly related to the `@Index` decorator, not just the `createIndexDecorator` function.

My solution was to use two fields, one with the correct name for the database (database field) and one virtual field for the code (application field) with magic get/set functions.
```
my_field = database field
myField = application field
```

```typescript
@Table({
  tableName: 'my_table',
[...]

PR #1192, opened 2022-01-30
Add support to underscored tables

GabrielNexT:
Hi guys, I was developing a small project and I found a bug.

Basically when we create an index with a table with the underscored property, we get an SQL error. As it seemed like a simple bug, I decided to explore to try to find a solution.

The idea I had is very simple, if you have the property, let's convert the name of all the fields of all the indexes to snake case and that's what I did, apparently it solves the problem.

This is my first contribution to an open source project, please be patient :)

Other people reported this bug in this issue #725

### DIFF
--- a/src/model/index/index-service.ts
+++ b/src/model/index/index-service.ts
@@ -16,10 +16,7 @@ export interface IndexesMeta {
   unnamed: IndexOptions[];
 }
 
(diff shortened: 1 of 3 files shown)


GalvinGao:
Any update on this? This looks like a very nice fix to merge into the upstream :D

PR #1205, opened 2022-02-11
build(deps-dev): bump dev dependencies

lukashroch:
Bump dev dependencies altogether as there were some of the packages required simultaneous bump, which dependabot was failing on.

Edit: lint actions on node 10 were turned off in CI. Eslint & markdownlint don't support node10 anymore. And it's not really needed to do linting on whole matrix, it's just dev step so should in fact be enough to run just one time.


Divlo (approved):
LGTM!

PR #1310, opened 2022-06-14
fix(model): compatible constructor with sequelize

lukashroch:
Makes the model constructor compatible with recent sequelize changes (https://github.com/sequelize/sequelize/commit/f5c06bd493670a37ba6d6ed039d44ccdf79b126e)


Divlo (approved):
LGTM! :+1:

PR #1331, opened 2022-06-19
chore(commitlint): disable body-max-line-length

WikiRik:
This is done so the auto-generated dependabot PRs won't fail the commitlint check


.commitlintrc.json
```diff
@@ -1 +1,6 @@
-{ "extends": ["@commitlint/config-conventional"] }
+{
+    "extends": ["@commitlint/config-conventional"],
+    "rules": {
+        "body-max-line-length": [0]
+    }
+}
```
ephys, line 6: Other JSON files in this project use 2 spaces per indentation level

ephys (commented):
A broader question about this: Do we want to align how PR are managed with the other Sequelize repos?

That would mean:
- Squash commits only
- Commit messages don't matter, only the PR title is linted (pending on our other repos too)

WikiRik:
Yes, that and things like the usage of dependabot vs renovate are things that we should discuss with the original team. We could take over fully and align our processes/styling with the main repo but I can imagine that not everybody will be that happy with that

lukashroch:
Alignment with sequelize sounds reasonable, imo it's not worth to fraction the workflows. Personally, I prefer renovate anyway. Though I like to group deps and let it run on weekly schedule or something ... these per-one-dependency commits just clutter the history too much for non critical deps 🙂

lukashroch (approved):

PR #1449, opened 2022-10-10
fix: globbing under Windows `models` option

vanthome:
The version 8 of the node-glob module only supports slashes.


skoniks:
> The version 8 of the node-glob module only supports slashes.

To make checks successful, i suggest to chage `.replaceAll` to `.split('\\').join('/')`

WikiRik:
Superseded by #1479

PR #1471, opened 2022-10-13
ci: sequelize renovate setup

lukashroch:
Sequelize-inspired renovate setup to sync the deps maintenance strategy. Additionally ignores couple of deps that drop earlier node.js support which sequelize V6 still retains. I think someone with enough rights will have to enable the renovate app for this repo if it is deemed to be merged.

@WikiRik, @sdepold, @ephys, what do you think? I'm hoping this should be easier to maintain same deps strategy as sequelize V6 and sequelize-ts@v2 will still linger around for a bit.

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,6 +1,24 @@
 {
   "extends": ["@commitlint/config-conventional"],
   "rules": {
-    "body-max-line-length": [0]
+    "body-max-line-length": [0],
+    "type-enum": [
+      2,
+      "always",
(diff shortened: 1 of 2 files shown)


lukashroch:
@WikiRik yeah, can probably be removed and use same convention sequelize is using to validate commit msgs. I only updated it to add the "meta" type sequelize is using.

WikiRik:
@sdepold @ephys I think one of you two needs to enable renovate btw

PR #1479, opened 2022-10-17
fix(deps): revert to glob@7.2.0 for sequelize@6 & node@10 compatibility

lukashroch:
reverting to  glob@7.2.0 for sequelize@6 & node@10 compatibility

new PR as GH didn't allow me to revert dependabots PR, not sure why :-/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "glob": "8.0.3"
+        "glob": "7.2.0"
       },
       "devDependencies": {
         "@commitlint/cli": "17.1.2",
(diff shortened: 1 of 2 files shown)


WikiRik:
The CI didn't pick this up before, should we do something about that? Add a test for the following function (the only usage of glob) maybe?
https://github.com/sequelize/sequelize-typescript/blob/f1354a4d7677c84fb21618536239c6ddeaab113f/src/sequelize/sequelize/sequelize-service.ts#L35-L73

WikiRik (approved):
I also wouldn't recommend reverting the Dependabot PR directly since more changes have occured to the package-lock file. Better to do this from scratch as you did now

lukashroch:
CI should probably have win platform, I think it's only ubuntu now. That would have picked this.

PR #1649, opened 2023-05-02
feat: added declare to docs

NGorco:


blankstar85:
I was just reading the updated sequelize docs and was thinking to add the pr to this and found this. I think you should put a blurb in the docs linking to the official docs about public fields so that users can understand the need for the change and the model reference. This would be a great pr to merge.

NGorco:
@blankstar85 Thanks, will do.

NGorco:
@blankstar85 Added more info

Issue #1195, opened 2022-02-02
How to get in touch regarding a security concern


JamieSlome:
Hey there!

I belong to an open source security research community, and a member (@sampaguitas) has found an issue, but doesn’t know the best way to disclose it.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

Sampaguitas:
Hey there,

I tried to get in touch with the developers to disclose a prototype pollution vulnerability in your sequelize-typescript npm package two years ago but never got a reply.

The latest version as of your npm package (2.5.1) is still vulnerable.

here is the [link to the security report posted on huntr.com platform](https://huntr.com/bounties/36a7ecbf-4d3d-462e-86a3-cda7b1ec64e2/)

I would appreciate if you validate it, a patch is also available to fix the vulnerability.

notes: this report is for the moment not public and can only be viewed by maintainers of the application.

WikiRik:
Thanks for sending another message. We've received it and will look into this issue. We will be giving updates on the huntr platform as well

Sampaguitas:
Thank you Rik

PR #1698, opened 2023-10-25
fix: deny modifying the object prototype

WikiRik:
cc @sampaguitas

Relates to #1195


sdepold (approved):
should this have a test?

WikiRik:
There is a proof of concept on the website they used to report this issue so we could adapt that one and use it

WikiRik:
Test added, not in the best file since `addScopeOptions` did not have any other tests and I didn't want to put it with the other scope tests since it will interfere with each other. Won't work on that more since we're deprecating sequelize-typescript with sequelize v7 either way

PR #1747, opened 2025-09-06
updated glob

deamondsear:


stemail23:
Hi, can I please understand the reticence to update the glob package? Cheers!

